PR #2, opened 2020-10-11
Change ethereumProviders to chains

andreogle:
Updates to the documentation for the recent changes to `chains` in `config.json`. I also tried to clean up this markdown file a bit

### DIFF
--- a/airnode/2-7-config-json.md
+++ b/airnode/2-7-config-json.md
@@ -37,7 +37,7 @@ Each node serves the APIs of a single provider.
 `triggers` are events that trigger an API call and an Ethereum transaction by Airnode.
 Triggers of different types are kept in lists under their respective keys:
 
-- `request`: When the node sees an event with its `providerId` and this trigger's `endpointId` emitted from the central Airnode contract, it responds to it with the respective endpoint defined in the OIS.
+- `request` - When the node sees an event with its `providerId` and this trigger's `endpointId` emitted from the central Airnode contract, it responds to it with the respective endpoint defined in the OIS.
   - `endpointId`
   - `oisTitle`
   - `endpointName`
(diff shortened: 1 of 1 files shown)


bbenligiray (approved):
These are not really comments on this specific PR but still made me think:

I think we'll need to generalize the node API-related fields (`nodeKey`, etc.) so that the user can create multiple API users (one for ChainAPI, another one for the open source node dashboard that they will run themselves for example). Mentioning this for the future.

What happens when the provider name isn't unique? I guess the user wouldn't be able to distinguish between the logs but it wouldn't error, right? Maybe non-unique provider names should be appended with indices or random strings at runtime, because I can see people copy pasting "ethereum" to different providers.

A default `minConfirmations` of 6 is a bit much. Chainlink uses 3. I think we can even go lower (0?) because we already wait an average of 30 seconds to see a request (because of the 1 minute cycles). I know this is not implemented yet btw.

andreogle:
1. Agree about generalizing the API-related `nodeSettings`

2. Yeah you're right about the logging. If you have 2 providers with the same name, the only effect would be logs where you couldn't distinguish the provider. It wouldn't break the node. I wanted to add an index suffix like `-0 and `-1` but it's tricky because you need to compare across nested arrays. i.e. provider names are stored in `chains` and then `providers`. You would need to collect all names at the start and then apply suffixes. It's definitely doable, I just didn't want to spend too much time on it yet. We can create an issue for this.

3. Yeah the `minConfirmations` defaulting to 6 sounds too long, I'm not exactly sure what would be best here. I initially thought a default of 1 might be nice, but on average there's a 1/4 chance a request a request arrives in the last 15 seconds of the cycle. Having to wait for 1 confirmation would then mean that these requests would likely need to wait an extra minute to be processed. Doesn't sound great to me. Not having to wait for any confirmations (at least by default) would be great

Issue #11, opened 2021-03-15
API key should be filled


drgy:
If I understood correctly API key in `config.json` has to be filled out or provided in `.env` file. This makes description [here](https://github.com/api3dao/api3-docs/blob/8e41ab2597cb19cb75e4904d6e9b214287c1d738/airnode/security-json.md#the-user-flow) deprecated, as the [related issue](https://github.com/api3dao/airnode/issues/163) was closed.

bbenligiray:
Fixed in https://github.com/api3dao/api3-docs/pull/12 and https://github.com/api3dao/api3-docs/pull/13

PR #12, opened 2021-03-17
Edit pre alpha

bbenligiray:
Some minor fixes + addresses https://github.com/api3dao/api3-docs/issues/11

I only updated the pre-alpha version but they should also be applied to the following versions where applicable


docs/pre-alpha/README.md
```diff
@@ -77,13 +77,15 @@ Not comfortable with academic-style papers? The following Blog Posts explain the
 
 1. [Airnode: The API gateway for blockchains](https://medium.com/api3/airnode-the-api-gateway-for-blockchains-8b07ff136840)
 2. [Getting to know Airnode](https://medium.com/api3/getting-to-know-airnode-162e50ea243e)
+3. [Setting Oracle Integration Standards](https://medium.com/api3/setting-oracle-integration-standards-ac9104c38f9e)
```
bbenligiray, line 80: Is there a way we can embed `docs/pages/medium.md` here?

docs/pre-alpha/README.md
```diff
@@ -11,30 +11,30 @@ title: What is API?
...
-Existing oracle solutions fall short because they fail to make this distinction, resulting in inferior solutions that depend on third-party oracles and ecosystems that exclude API providers. By refining the definition of the problem, **API3** aims to provide a much more optimal solution.
+Existing oracle solutions fall short because they fail to make this distinction, resulting in inferior solutions that depend on third-party oracles and ecosystems that exclude _API providers_. By refining the definition of the problem, **API3** aims to provide a much more optimal solution.
 
 ::: tip API3 Whitepaper (pdf)
 For a detailed discussion of the API3 project, read the [API3 Whitepaper](https://github.com/api3dao/api3-whitepaper/blob/master/api3-whitepaper.pdf).
 :::
 
 ## The Airnode
 
-At its core, **API3** brings the ability for API Providers to easily run their own Oracle nodes. This allows them to provide their data on-chain, without an intermediary, to any Decentralized App \(dApp\) interested in their services.
```
bbenligiray, line 30: In general we should never capitalize oracle because it is already easy to confuse with Oracle the company

docs/pre-alpha/airnode/specifications/reserved-parameters.md
```diff
@@ -11,8 +11,8 @@ A requester can pass request parameters either by referencing a [template](../..
 In either case, these parameters are encoded in a `bytes`-type variable using [Airnode ABI](airnode-abi-specifications.md).
 There are two types of parameters:
 
-1. [Endpoint parameters](https://github.com/api3dao/api3-docs/blob/master/airnode/ois.md#55-parameters) mapped to API operation parameters
-1. [Reserved parameters](https://github.com/api3dao/api3-docs/blob/master/airnode/ois.md#54-reservedparameters)
+1. [Endpoint parameters](https://api3dao.github.io/api3-docs/pre-alpha/airnode/specifications/ois.html#_5-5-parameters) mapped to API operation parameters
```
bbenligiray, line 14: We should use relative links here

docs/pre-alpha/airnode/specifications/security-json.md
```diff
@@ -45,6 +45,3 @@ Here is an example `security.json` file that is generated by a third party (e.g.
   "id": "9a52131e-babd-11ea-b3de-0242ac130004"
 }
 ```
-
-The first option here is for the user to open `security.json` and replace `<TO_BE_FILLED>` with their API key.
```
bbenligiray, line 49: https://github.com/api3dao/api3-docs/issues/11

docs/pre-alpha/tutorials/airnode-starter.md
```diff
@@ -56,7 +58,7 @@ However, if you will be deploying your own Airnode, the provider endpoint must b
 
 *(You only need cloud credentials if you will not be skipping Step 1.)*
 
-Follow the [docs](https://github.com/api3dao/api3-docs/blob/master/provider-guides/deploying-airnode.md#creating-cloud-credentials) to create your cloud credentials.
+Follow the [docs](https://api3dao.github.io/api3-docs/pre-alpha/guides/provider/deploying-airnode.html#creating-cloud-credentials) to create your cloud credentials.
```
bbenligiray, line 61: Again, relative links if possible

bbenligiray:
After this is merged and deployed, let's add a Docs tab to the website.

docs/pre-alpha/README.md
wkande: Do you mean to embed the markdown file medium.md into README.md as opposed to it having its own medium content as it currently does? Embedding an MD into an MD is messy. Markdown does not support an "include". I could look into writing a component. Might take some thought.

bbenligiray: Yeah I just wondered if there is an easy way of doing it. Regardless, I think it's a bad idea to try to maintain two separate lists of Medium posts so we need a different solution (for example remove the one in `README.md` and link to `pages/medium.md` instead).

wkande: A link to medium.md from README.md would result in the sidebar being removed by Vuepress because we would leave the version's root folder of /pre-alpha. The idea behind the folder **/pages** is that VuePress uses it for common files but without sidebar support. 

I will have to think more about it. Right now all I can come up with is a custom component that lists the medium posts. The component can then be embedded into any MD file. Actually should be easy to write and maintain. Instead of markdown links the component would use HTML links. I will give it a try.

wkande:
> I only updated the pre-alpha version but they should also be applied to the following versions where applicable

I see that in fact, you updated some version **next** files as well. I can move the pre-alpha changes to **next** after the merge.

wkande (approved):
Will merge these and follow through on the pre-alpha changes that need to go to "next".

PR #28, opened 2021-03-27
Fix several broken links in the README

gwn:
Not really the biggest problem as there are no issues in the deployed documentation, but there are people like me browsing it from Git browsers too, so it'd be nice if the links are not broken :)


wkande:
Thanks for getting this done.

PR #46, opened 2021-04-06
Hyperlink config.json fields

bbenligiray:
It's a bit redundant with the TOC


wkande:
I see you have place links on all heading2 elements pointing back to themselves. This just pushes the heading to the top of the page. Is that what you wanted?

bbenligiray:
I don't think I did, can you take another look?

PR #110, opened 2021-06-29
DAO Docs -- first draft

EzraWeller:
It's late so I may have forgotten something, but I think this is an okay first draft. Let's do a round of feedback and edits / additions, @bbenligiray .


umersin61:
The explanation for depositRegular(uint256 amount) is the explanation for delegateVotingPower(address delegate). 

image

umersin61:
Something about the lock on those minted tokens would be nice here.
image

umersin61:
Picture seems to be off in api3-docs/docs/pre-alpha/members/governance/proposals.md
image

EzraWeller:
> Something about the lock on those minted tokens would be nice here.
> image

Totally. I actually just found out about this. Could you or someone explain point me to how this lock works? I didn't notice it in the contracts or in the public test on Rinkeby.

umersin61:
According to the contract it should be function 2, so 52 Epochs (which are currently 1 week long).
As far as i am aware the reward vesting period can also be adjusted by the DAO.
image

bbenligiray:
> According to the contract it should be function 2, so 52 Epochs (which are currently 1 week long).
> As far as i am aware the reward vesting period can also be adjusted by the DAO.
> image

This is not really a function but a public variable. Although these are callable like functions, I think it's fine to not document them as a part of the interface just to avoid crowding things up. Also it's constant, can't be updated.

EzraWeller:
Okay, info on locking should now be added. Thanks for the info. 

Let me know if anyone is still having trouble with pages showing up blank / images not loading -- I'm still not seeing any issues there.

docs/pre-alpha/members/contract-architecture/pool.md
```diff
@@ -0,0 +1,44 @@
...
+|`depositRegular(uint256 amount)` |Deposits your API3 tokens into the pool. Tokens must be deposited before they can be staked. |
+|`stake(uint256 amount)`|Stakes deposited API3 tokens. Staked tokens will earn rewards, grant voting power (and may be slashed if there is a claim on the pool—not yet implemented). |
+|`depositAndStake(address source, uint256 amount)` |Deposits and stakes API3 tokens in one transaction. |
+|`scheduleUnstake(uint256 shares)` |Schedules staked tokens to be unstaked. In order to unstake API3 tokens, members must first schedule an unstake and wait the scheduled period before unstaking (currently ~1 week). Tokens scheduled to be unstaked no longer grant voting power or earn rewards.  |
+|`unstake(address userAddress) returns(uint256)` |Unstakes API3 tokens, allowing them to be withdrawn from the pool (unstaking and withdrawing are separate steps). Can only be called after scheduling an unstake and waiting the scheduled amount of time. |
+|`withdrawRegular(uint256 amount)` |Withdraws your API3 tokens from the Pool contract. |
+|`unstakeAndWithdraw(address destination)` |Unstakes and withdraws tokens in one transaction. |
+
+### Timelock Functions
```
acenolaza, line 25: Since we are mentioning the "Timelock Manager contract" then I think it could be useful to add it to the diagram or a note about this contract in [here](https://github.com/api3dao/api3-docs/blob/58a8cbc27c4216174970d66f8e5da512e51cc869/docs/pre-alpha/members/contract-architecture/overview.md).

docs/pre-alpha/members/contract-architecture/pool.md
```diff
@@ -0,0 +1,44 @@
...
+|--- |--- |
+|`function deposit(address source, uint256 amount, address userAddress)` |Callable only by the Timelock Manager contract. Deposits tokens on behalf of a user. |
+|`function depositWithVesting(address source, uint256 amount, address userAddress, uint256 releaseStart, uint256 releaseEnd)` |Callable only by the Timelock Manager contract. Deposits tokens on a vesting schedule on behalf of a user. These vesting tokens cannot be withdrawn by the user until they have vested, but they *can* be staked. |
+|`function updateTimelockStatus(address userAddress, address timelockManagerAddress)` |Updates the vesting status of a user's deposited vesting tokens (i.e. unlocks tokens) according to the schedule in the Timelock Manager contract. |
+
+### Voting Power
+|Signature | Description|
+|--- |--- |
+|`userVotingPowerAt(address userAddress, uint256 _block)`|Returns a user's current voting power (0 if they have delegated it). |
+|`delegateVotingPower(address delegate)` |Delegates a member's voting power, as decided by their share of the staking pool, to another address. |
```
acenolaza: We could also add that this function can be used to update the delegation to another user to avoid unnecessary txs by users calling undelegate first if they are just trying to deletegate their voting power to another user.

docs/pre-alpha/members/contract-architecture/pool.md
```diff
@@ -0,0 +1,44 @@
...
+|`undelegateVotingPower()` |Undelegates a member's voting power. |
+
+### Other
+|Signature | Description|
+|--- |--- |
+|`setProposalVotingPowerThreshold(uint256 _proposalVotingPowerThreshold)` |Sets the minimum voting power required for a member to create a new proposal. Can only be adjusted by the primary voting app. |
```
acenolaza: If we have this function here shouldn't we also add other governance configuration functions like: setProposalVotingPowerThreshold, setAprUpdateStep, setUnstakeWaitPeriod, setStakeTarget, etc? I know the idea is to keep it as simple as possible but this could make it easier for the users when creating proposals.

acenolaza (commented):
nit-pick comments but I think it would be nice to add an example of a proposal since it's the most complicated process in the dao.

docs/pre-alpha/members/contract-architecture/pool.md
EzraWeller: Good point -- this is already listed in the proposals guide, so we can remove it here I think.

docs/pre-alpha/members/contract-architecture/pool.md
EzraWeller: Hmm, I wanted to keep the diagram relatively simple. If we add the timelock manager, I think we may also have to add the token contract. Is it getting too complex at that point?

EzraWeller:
> nit-pick comments but I think it would be nice to add an example of a proposal since it's the most complicated process in the dao.

There is a simple example in a screenshot on the proposal page, does that work?

docs/pre-alpha/members/contract-architecture/pool.md
acenolaza: I thought the same and I was wondering if I should also suggest adding the token contract to the diagram but if the idea is to keep it small and simple then leave it as is.

acenolaza:
> > nit-pick comments but I think it would be nice to add an example of a proposal since it's the most complicated process in the dao.
> 
> There is a simple example in a screenshot on the proposal page, does that work?

hmmm, weird, I do see content in the proposals.md and voting.md files but when I run the docs locally those pages are empty. I must be doing something wrong.

EzraWeller:
Hmm, I think Burak was having a similar problem in his local env. Not sure what's going on there, any ideas how to figure it out?

acenolaza:
> Hmm, I think Burak was having a similar problem in his local env. Not sure what's going on there, any ideas how to figure it out?

Ok. I think I've found the root cause of the problem. It was not building due to (vuepress?) not being able to find images since the extension was uppercase `PNG` and it was looking for files with lowercase `png`.  I just renamed the files and everything is back to working (even proposals and voting page). I haven't look for an option to make this case insensitive but if you want me to push the file renames then please let me know. In the meantime I'll review those pages that I'd missed.

EzraWeller:
Ah, thanks windows snip tool for that one. Feel free to change it or I'll get to it when I can soon.

wkande (approved):
This looks really good everyone.

wkande:
I resolved one conflict with the pre-alpha sidebar.js, nothing major.

docs/pre-alpha/members/governance/proposals.md
```diff
@@ -0,0 +1,41 @@
...
+## Proposal Creation
+Proposals can be created from the "Governance" tab of the DAO portal. Connect an account with sufficient voting power and click "New Proposal." Then fill out the form. Note the tooltips and the formatting in this example:
+
+<p align="center">
+  <img src="../../figures/new-proposal.png" width="700" />
+</p>
+
+Proposals can be submitted to either the *Primary* or *Secondary* voting app. These two apps have access to separate treasuries, have different voting settings, and have different permissions to change contract settings. 
+
+In general, the Primary voting app has a larger treasury and more permissions but has more stringent voting settings. For a technical breakdown of the different permissions granted to the DAO's two voting apps (and corresponding Agents) see this [README](https://github.com/api3dao/api3-dao/blob/develop/packages/dao/README.md#permissions).
```
acenolaza, line 22: Shouldn't the README url point to the main branch instead of develop? https://github.com/api3dao/api3-dao/blob/main/packages/dao/README.md

PR #162, opened 2021-08-18
Update designated-wallet.md

ErichDylus:
Clarifying the control language on designated/sponsored wallets, to alleviate confusion on the 'custodial' term as there is no custody obligation on excess funds to the provider or otherwise imposed due to the de minimus nature and lack of express or implied obligation from the provider.


wkande:
I took the text from the PR and moved it into a Vue component. Since /pre-alpha uses designated wallet and /next will use sponsor wallet, I actually created two components. 

Since Burack had mentioned before that we should show this “warning” in more than one strategic location for each version of the docs the components make sense. Now if you wish to change the text you can do so from a single location. Just do a PR on the component. You will notice it is not markdown but rather a simple HTML.

The following files show the two components now in use on the docs website. Also, I have closed the PR without merging.

/pre-alpha
designated-wallet.md 
creating-a-requester.md 

/next
sponsor-wallet.md 
developers/README.md 
developers/sponsorships.md
developers/fees.md

PR #169, opened 2021-08-21
Fix minor typo

dcroote:


wkande:
Thanks again for the catch.

PR #176, opened 2021-08-27
Configuration changes regarding secrets interpolation

amarthadan:
This is (mostly) regarding [AN-73](https://api3dao.atlassian.net/browse/AN-73). You don't have to merge this but it felt better to show you what has changed this way in comparison with the old version instead of just a separate document.


wkande (approved):
This is a great PR saving me a ton of work and teaching me what changed. Thanks

wkande:
Again this is a real big help to say the least. From here I will be able to change the file configuring-airnode.md with ease or at least with a lot fewer questions.

PR #184, opened 2021-09-01
Add redirects for API Readme links

37Rb:


wkande (commented):
I am assuming you have a use case for these links but wanted to be sure you understand the use of **/next**. /next will most likely become **/0.1**

So when we go to production the following:
`/d/call-an-airnode      /next/grp-developers/call-an-airnode.html`
will become:
`/d/call-an-airnode       /0.1/grp-developers/call-an-airnode.html`

And there is no guarantee the the path part of the redirect will remain the same. Some files are still in the process of being changed and they might get moved around.

The new release of Airnode is not production ready so remember that the docs under **/next** are describing a product that is not available.

Anyway remember that these redirects will need to be fixed (not the route) when we release **0.1**.

37Rb:
I understand. The initial use case for these redirects for Airnode API docs. For example:

https://github.com/37Rb/sample-airnode-doc

Although I think we will eventually want a permanent URL for most high level concepts that we can link to from the website and documentation. But I can create them as needed. We don't have to try to anticipate them now.

If you want we can change them to link to pre-alpha instead of next. All I really care about is that the routes never change.

wkande (approved):
Merging the PR. The routes are guaranteed not to change but for paths with **/next**, the path will change when **/next** goes to production. All paths in the redirect file are updated manually as part of the versioning process, see [Versioning](https://docs.api3.org/dev/versioning.html)

PR #187, opened 2021-09-02
AN-178: Return dummy chain metadata on pre-alpha instead of leaving i…

acenolaza:
…t undefined

It was decided that safest option in **pre-alpha** was to disable chain specific metadata information since Airnode.sol does not emit chainId in the events after requests are submitted.

@wkande is main branch the right base branch?


wkande:
Looks simple enough, thanks Santiago.

PR #196, opened 2021-09-06
Remove outdated convertor section

Siegrift:
In the https://docs.api3.org/next/grp-providers/guides/provider/api-integration.html there is a text `Currently there is no tool that converts OAS to OIS automatically. If you would like to help build this, pleas...`  which is outdated. 

That issue is closed and converter is implemented (at least partially). I'd remove that section for now and add the converter section later?


wkande (approved):
Looks good, I have additional changes for this page as well but will also add a marker about the need to document the converter.

PR #201, opened 2021-09-07
Update the community nav links

37Rb:
We're deemphasizing TG as a place to send people and promoting our owned channels such as the forum and discord.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ The API connectivity problem is in essence an ecosystem building problem, more s
 
 **Website:** https://api3.org/
 
-**Telegram (for community):** https://t.me/API3DAO
+**Community Forum:** https://forum.api3.org/
 
-**Discord (for contributors):** https://discord.gg/qnRrcfnm5W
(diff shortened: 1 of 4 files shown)


wkande:
@37Rb So Telegram is going away? If so there are three more references to Telegram in these docs. Should we replace them as well?

I do seem get a Telegram notification almost every day from API3. Maybe we keep Telegram in the docs below and add the Forum?

- root README.md
- docs/next/README.md
- docs/pre-alpha/README.md

37Rb:
Telegram isn't going away but we don't want to send people there. I'll update those references as well. Thanks!

wkande:
I have updated the references in the README files so no need to worry.

wkande:
I had already merged a branch of mine with the Forum added to the (3) READMEs. Also I had another branch that was full of changes for the sidebar (navigation changes) so I added in the navbar change of yours. This way I was able to avoid merge conflicts between our 3 PRs.

Everything should be there now. I am closing this PR without merging.

PR #205, opened 2021-09-08
Fix style font-size typo

dcroote:


wkande:
Thanks Derek for the catch.

PR #212, opened 2021-09-12
Add docs section and adjust links in Contributing

dcroote:


wkande:
And thanks again This makes sense.

PR #214, opened 2021-09-12
Update Proposals language and misc

dcroote:
The PR updates the following on the Proposals page:

- Typo: `proposalVotingPowerThreshold` was erroneously `1 * 10^17` which equals 10%. I've updated it to `0.1 * 10^16` so it is consistent with the other 10^16 values (although it could also be 1*10^15)
- Stated explicitly `10^18 = 100%`
- Internal & external links
- Grammar
- Consistent use of "DAO Dashboard"
- Rephrased sentences within the ENS section to encourage use and more clearly illustrate setup for raw address vs. multisig

### DIFF
--- a/docs/pre-alpha/members/governance/proposals.md
+++ b/docs/pre-alpha/members/governance/proposals.md
@@ -8,10 +8,10 @@ title: Proposals
 <TOC class="table-of-contents" :include-level="[2,3]" />
 
 ## Requirements
-To create a proposal, you must hold at least 0.1% of the total pool shares. This required percentage is set in the Pool contract can be adjusted by the DAO.
+To create a proposal, you must hold at least 0.1% of the total pool shares. This required percentage is set in the Pool contract and may be updated by a DAO proposal as described under [Proposal Types](#proposal-types) below.
 
 ## Proposal Creation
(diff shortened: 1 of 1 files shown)


wkande:
To be honest I would have no idea what the right calculations might be. If you are confident then we are good on that.

wkande:
I also agree with you about calling the DAO dashboard "DAO Dashboard" but there are differences of opinion on that. For instance I want to call an API provider "API Provider" but again there is not a general consensus on doing so.

I have been doing so in the sidebar (and Heading elements) as it just looks better. Not sure everyone agrees. I am still trying to sell it. We should have a discussion on this to create a common opinion.

dcroote:
> If you are confident then we are good on that.

Yes I'm confident it should be `0.1 * 10^16`. Here is where `ONE_PERCENT = 1e18 / 100`: 

https://github.com/api3dao/api3-dao/blob/c745454d60ad97f6ffa9b5aa612c9a2b7eba16d0/packages/pool/contracts/StateUtils.sol#L61-L62

And here is where `proposalVotingPowerThreshold` is set to 0.1%:

https://github.com/api3dao/api3-dao/blob/c745454d60ad97f6ffa9b5aa612c9a2b7eba16d0/packages/pool/contracts/StateUtils.sol#L153
<br /><br />

> I have been doing so in the sidebar (and Heading elements) as it just looks better

I'm happy to align on whatever phrasing for DAO Dashboard, I just wanted to make it consistent with what I saw in the sidebar

PR #216, opened 2021-09-13
Add more detail about http-server install

### LINKED ISSUE #215: Add http-server to devDependencies in package.json?
While not strictly necessary to run `yarn docs:dev` locally, building the docs to run [link validation](https://docs.api3.org/dev/link-validator.html#step-2-start-http-server) requires it.

wkande:


wkande:
I added info about install http-server globally or as a dev dependency. Either should work. Since many will already have it install globally, `npm install http-server --save-dev` is an optional usage.

PR #202, opened 2021-09-07
Post-processing docs

bbenligiray:

PR #223, opened 2021-09-14
Revert "Post-processing docs"

wkande:
Reverts api3dao/api3-docs#202

So this will (I hope) revert PR #202 for the post-processing changes. @bbenligiray had changed ois.md and added post-processing.md. No on has touched either file since the #202 was merged so it should go through just fine without conflicts.


wkande:
@bbenligiray Can you review this? I'm sure this is what you wanted to do with your post-processing changes.

wkande:
@bbenligiray I will revert this then. Your old branch will be retained. In addition I had place `osi.md` and `post-processing.md` from the branch into the _hold directory of the project just in case.

PR #233, opened 2021-09-22
AN-211

wkande:
A massive change to the layout of the docs. A few new components and changes to others.


wkande:
The link check failed because `/pre-alpha/ has moved to `/airnode`.

PR #240, opened 2021-09-23
Prep work for Authorizers AN-198

wkande:


wkande:
Two links failed the check but will get corrected in PR237

PR #241, opened 2021-09-23
Fix broken symlink causing CI failure

dcroote:


wkande:
@dcroote Merging this PR now.

PR #243, opened 2021-09-24
AN-120: Document adminnable contracts

acenolaza:
Added a new adminnable.md file under Concepts and Definitions. Also included a small description of the Whitelister contract


wkande (approved):
This is helpful thanks

PR #264, opened 2021-10-01
Fixed the icon flickering when the SPA loads

wkande:


wkande:
JIRA issue AN-249.

Mostly on Firefox, when the SPA loads the button in the component flash themselves very large in the sidebar. Probably need to hide them until the mounted function fires so they draw the right size when first shown.

PR #270, opened 2021-10-03
Mention that HTTP gateway is only available on serverless

Siegrift:
I think we should mention features which are only available on a particular "airnode type" explicitely.

But you can close this if you disagree.


wkande (approved):
Good catch, thanks.

PR #273, opened 2021-10-04
Stop linking pre-alpha requester examples, link to examples in airnode repo instead

Siegrift:
https://api3dao.atlassian.net/browse/AN-189


wkande (commented):
Not sure we are just linking to the newer airnode-starter (examples). My understanding was that this would evolve into a tutorial in the Docs.

Also the links to the new requester contracts are pointed to a branch and not the master.

Siegrift:
> Not sure we are just linking to the newer airnode-starter (examples). My understanding was that this would evolve into a tutorial in the Docs.

It should at some point, but I think linking is sufficient for now. We have issues for migrating READMEs to docs, but we didn't include those in v0.1.0.

> Also the links to the new requester contracts are pointed to a branch and not the master.

Yes, that's why I created a draft only. I am waiting until that is merged and will change the links afterwards.

wkande:
> We have issues for migrating READMEs to docs, but we didn't include those in v0.1.0.

Not clear here. Are you saying that v0.1.0 will not have airnode-starter in the Docs but we will link to the the repo README in the examples package?

Siegrift:
Yes. We have these issues created in the jira: 
https://api3dao.atlassian.net/jira/software/projects/AN/boards/18/backlog?text=readme

we can add another one to migrate the examples README.

wkande:
I do recall those but thought airnode-starter would not be inclusive. We should probably bring this up at the Wednesday meeting. I think that the airnode-starter is one of the most important guides for new users. Derek has been tasked to bring the concepts of the README into the Docs and enhance the content into a formal guide.

wkande (approved):
These look good.

PR #287, opened 2021-10-08
Add prettier configuration

Siegrift:
I've added prettier support (and a script to package.json) which formats the files in the project.

This works mostly fine, but it treats the files as `md` files and doesn't understand some of the vuepress specific stuff (e.g. tabs). 
However, I didn't find any vue specific feature which would be broken by the formatting (you mostly need to add newlines somewhere).

Even if there was some section that should not be formatted, you can exclude it from prettier: https://prettier.io/docs/en/ignore.html

I didn't format anything just yet not to create any conflicts for you. We can reformat the stuff gradually. (This also means you shouldn't run the `format` script, because that tries to format all files in the project).

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,3 @@
+/node_modules
+/archived
+/docs/airnode/pre-alpha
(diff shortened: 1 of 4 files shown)


wkande:
We should take time to go over this and decide how to use it.

PR #292, opened 2021-10-09
Client image docs

Siegrift:
https://api3dao.atlassian.net/browse/AN-268

### DIFF
--- a/docs/airnode/next/grp-providers/docker/client-image.md
+++ b/docs/airnode/next/grp-providers/docker/client-image.md
@@ -1,5 +1,5 @@
 ---
-title: Client Image
+title: Airnode Client Image
 ---
 
 <TitleSpan>Docker Images</TitleSpan>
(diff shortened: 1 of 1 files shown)


wkande (commented):
This looks really good to me. I see mention of using Windows powershell. In other docs, deployer image, we mention the user should use cmd and not powershell for the deploy command. See [here](https://docs.api3.org/airnode/next/grp-providers/docker/deployer-image.html#deploy).

Siegrift:
> I see mention of using Windows powershell. In other docs, deployer image, we mention the user should use cmd and not powershell for the deploy command. See here.

Oh, I didn't notice that we instruct them to use CWD - but I find that a silly requirement. We should probably change the other usages (but that's a low prio)

wkande:
Ok then. I will just go ahead and merge this. That is something we can work out latter. I will make a GitHub issue for it and move it to JIRA if there is a decision one way or another.

PR #305, opened 2021-10-13
Update admin CLI doc

Siegrift:
You can review this commit by commit (since the first commit just formats the document).

PR #307, opened 2021-10-13
Update the Airnode concept doc

Siegrift:
I need https://github.com/api3dao/api3-docs/pull/305 to be merged first, because there is one link I want to add which is only avaialble after that PR is merged.


dcroote:
@Siegrift - with #305 merged it looks like this is now unblocked per your comment.

PR #310, opened 2021-10-14
Add airnode protocol deployment contract addresses

dcroote:
AN-236

### DIFF
--- a/docs/airnode/next/reference/airnode-addresses.md
+++ b/docs/airnode/next/reference/airnode-addresses.md
@@ -7,24 +7,22 @@ title: Airnode Contract Addresses
 <TocHeader />
 <TOC class="table-of-contents" :include-level="[2,3]" />
 
-This doc shows EVM-compatible blockchains that Airnode protocol contracts have been deployed to. Use these contract addresses to interact with Airnode.
-
-<Fix>This doc will replace the Vue component that also has the contract addresses. That component is not practical when used in other docs and can distract the user's focus of the local content.</Fix>
+Use the contract addresses listed in the tables below to interact with Airnode on EVM-compatible blockchains. Additional addresses will be added as contracts are deployed, but feel free to [submit a GitHub issue](https://github.com/api3dao/airnode/issues) requesting a new deployment.
 
 ## AirnodeRrp.sol
 
 | Blockchain | ID  | Address                                    |
 | ---------- | --- | ------------------------------------------ |
-| mainnet    | 1   | ...                                        |
+| mainnet    | 1   |                                            |
 | ropsten    | 3   | 0xF6d2675468989387e96127546e0CBC9A384fa418 |
-| rinkeby    | 4   |                                            |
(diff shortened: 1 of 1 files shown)


docs/airnode/next/reference/airnode-addresses.md
```diff
@@ -7,24 +7,22 @@ title: Airnode Contract Addresses
...
 ## AirnodeRrp.sol
 
 | Blockchain | ID  | Address                                    |
 | ---------- | --- | ------------------------------------------ |
-| mainnet    | 1   | ...                                        |
+| mainnet    | 1   |                                            |
 | ropsten    | 3   | 0xF6d2675468989387e96127546e0CBC9A384fa418 |
-| rinkeby    | 4   |                                            |
+| rinkeby    | 4   | 0xe469aD70a2CD3A01F4eCb6502161A4F381c1F659 |
 | kovan      | 62  |                                            |
```
Siegrift, line 19: I wonder wheter we shall not remove kovan (or deploy the contracts also there).

docs/airnode/next/reference/airnode-addresses.md
```diff
@@ -7,24 +7,22 @@ title: Airnode Contract Addresses
...
 ## AirnodeRequesterRrpAuthorizer.sol
 
 | Blockchain | ID  | Address                                    |
 | ---------- | --- | ------------------------------------------ |
-| mainnet    | ... | ...                                        |
-| ropsten    | 3   | ...                                        |
-| rinkeby    | 4   | ...                                        |
-| kovan      | 62  | ...                                        |
+| mainnet    | 1   |                                            |
+| ropsten    | 3   |                                            |
```
Siegrift, line 26: Can we deploy this on ropsten as well?

Siegrift (approved):
LGTM, but maybe it will look better if there are no gaps?

But, we will revisit this doc once more when we deploy to mainnet, so we can address that at that point.

docs/airnode/next/reference/airnode-addresses.md
dcroote: Perhaps remove kovan for now and if someone wants it they'll open an issue?

docs/airnode/next/reference/airnode-addresses.md
dcroote: It may be valuable to offer the option to use ropsten or rinkeby, if deployment is not too much effort. Not sure I'm familiar enough with the `@api3/protocol` deployments for it though.

docs/airnode/next/reference/airnode-addresses.md
Siegrift: Yeah, I'd say so.

docs/airnode/next/reference/airnode-addresses.md
Siegrift: Let's do that when we deploy the contracts on mainnet.

PR #318, opened 2021-10-15
Adds more links to doc references and minor improvements

acenolaza:
Checked all commands and added more links to other parts of the docs.


wkande (commented):
@acenolaza - I went through this commit and AN-180 and AN-181. If I am reading this right both JIRs are addressed here?

acenolaza:
> @acenolaza - I went through this commit and AN-180 and AN-181. If I am reading this right both JIRs are addressed here?

Yes. Actually they were both addressed already and I improved this file by adding more links and better organized each section. I also updated some terms and added more details on some commands.

PR #334, opened 2021-10-19
Remove unused `ois.[0].apiSpecifications.security` field

amarthadan:
This was already removed in https://github.com/api3dao/api3-docs/commit/e874b2795c69a6845b0dfe00c59ec003b6072f08, not sure why it came back :smile:

### DIFF
--- a/docs/airnode/v1.0/reference/examples/config-json.md
+++ b/docs/airnode/v1.0/reference/examples/config-json.md
@@ -92,12 +92,7 @@ title: config.json
               "name": "X_access_key"
             }
           }
-        },
-        "security": [
-          {
-            "Currency Converter Security Scheme": []
-          }
-        ]
+        }
       },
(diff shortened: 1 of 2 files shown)


wkande:
@amarthadan I had just added the security field back in because Jakub told me it needs to be there.

amarthadan:
> @amarthadan I had just added the security field back in because Jakub told me it needs to be there.

That must have been some miscommunication. I've checked and it's not used anywhere in the code so it really shouldn't be there.

drgy:
I think there was a misunderstanding, the `security` was missing in one list in the docs, I assumed it was supposed to be there, as the `security` parameter was described on multiple places in the docs, I have no idea what the node itself needs to run correctly.

wkande:
Sounds good then. I will merge this and double check it to be sure it is gone completely, Thanks

PR #337, opened 2021-10-19
Add why use airnode docs page

Siegrift:
This is a page I (subjectively) had in mind and was missing.

I realize this is subjective (and I am not a particularly experienced writer) so I'd like feedback and proofread.
Any thoughts what could be added?

### DIFF
--- a/docs/airnode/v1.0/sidebar.js
+++ b/docs/airnode/v1.0/sidebar.js
@@ -3,7 +3,7 @@ module.exports = [
     title: 'Introduction', initialOpenGroupIndex: 0, collapsable: true,
     children: [
       '',
-      //'introduction/why-use-airnode'
+      'introduction/why-use-airnode'
     ]
   },
   {
--- a/docs/common/why-use-airnode.md
+++ b/docs/common/why-use-airnode.md
@@ -6,50 +6,47 @@ title: Why use Airnode?
 
 <TOC class="table-of-contents" :include-level="[2,3]" />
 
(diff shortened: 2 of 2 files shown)


wkande:
@Siegrift  So I like the **pros** list. The doc overall quickly answers it own question "Why use Airnode?" Not so sure about the **cons**. This doc will be one of the very first that new users will read. Its very title will draw them in. From a marketing perspective the cons could have a negative impact. Should we comment out the **cons** and see if later they might be better discussed somewhere else?

Siegrift:
> Should we comment out the cons and see if later they might be better discussed somewhere else?

Nope, it should get discussed here and now. If it's something we don't want to have I'll remove it.

wkande:
OK why don't we go with it and get input (or see what the reactions are). I might go back and try and candy coat them a little.

Siegrift:
Yeah, there are assigned as reviewers on the PR, so they already received a notification :)

amarthadan (commented):
I personally quite like when some software documentation has a cons section. It feels more like they know what they're doing, these are the compromises that needed to be made and they're giving you all the information. So, I'm for keeping that section even without any sugarcoating, but if you think it's necessary I'm ok with that.

wkande (approved):
I can live with this.

dcroote:
@Siegrift  - I've updated the language, softening parts but remaining fair and retaining the key messages elsewhere. There ended up being enough edits that suggesting changes through github wasn't productive so I pushed directly.

Siegrift:
I agree this is much better. I'm merging.

PR #338, opened 2021-10-19
Fulfill diagram update

acenolaza:


wkande (approved):
An improvement on the diagram text to be sure.

PR #349, opened 2021-10-22
Update admin CLI mnemonic docs

Siegrift:
https://api3dao.atlassian.net/browse/AN-289

### DIFF
--- a/docs/airnode/v0.2/reference/admin-cli-commands.md
+++ b/docs/airnode/v0.2/reference/admin-cli-commands.md
@@ -97,7 +97,7 @@ Commands related to a [sponsor's](../concepts/sponsor.md) relationships between 
 Sponsoring a requester and using the returned `sponsorAddress` to derive a `sponsorWallet` for an Airnode creates a [relationship](../concepts/sponsor.md) between the requester and the Airnode, see the [`derive-sponsor-wallet-address`](admin-cli-commands.md#derive-sponsor-wallet-address) command.
 
 - `provider-url`: A valid blockchain provider URL.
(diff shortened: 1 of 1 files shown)


docs/airnode/v0.2/reference/admin-cli-commands.md
```diff
@@ -171,7 +171,7 @@ npx @api3/airnode-admin derive-sponsor-wallet-address \
 Reads a file, uses its contents to create a [template](../concepts/template.md) and returns a `template-id`. Also see [Using Templates](../grp-developers/using-templates.md) for an example template file.
 
 - `provider-url`: A valid blockchain provider URL.
-- `mnemonic`: A wallet owned by the sponsor. Used to pay gas costs from the mnemonic's default account unless a `derivation-path` is specified.
+- `mnemonic`: Used to pay gas costs from the mnemonic's default account unless a `derivation-path` is specified.
```
Siegrift, line 174: This was wrong previously (I strongly believe)

wkande: So the part that you think is wrong was "A wallet owned by the sponsor." Who, other than a sponsor, might create a template? I'm thinking an API provider creates one for requesters to use against their APIs.

Siegrift: technically anoyone (there is no limitaiton). But I guess most of the time it will be created by API provider as you're saying.

wkande: I'm wondering if that need to be mentioned in the command's description?

Siegrift: I don't think so.

wkande (approved):
This is good to go. I am merging this now along with Derek's PR.

PR #353, opened 2021-10-23
Rename packages

Siegrift:
https://api3dao.atlassian.net/browse/AN-302


wkande (approved):
Looks like this PR gets merged first before before the other one? It has conflicts with your other `.prettierignore` file. How about you merge the two PRs when you are ready. From what I can tell your plan sounds fine.

PR #357, opened 2021-10-24
Add scrollbar to stop navbar shift on short pages

dcroote:
This PR adds a permanent scrollbar so navigating between pages with less than and more than a screenful of content does not shift the top navigation bar. Tested in Chrome, Brave, Firefox, Edge.

Short page without scrollbar, orange line for reference:
image

Longer page with scrollbar, orange line in same place shows how the navbar shifts because of the scrollbar:
image

### DIFF
--- a/docs/.vuepress/styles/index.styl
+++ b/docs/.vuepress/styles/index.styl
@@ -40,6 +40,11 @@
   border-left:solid #7963B2 5px !important;
 }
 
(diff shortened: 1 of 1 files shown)


wkande (commented):
Can you place a code comment on the css as well so down the road we have reference as to why we changed the html element?

dcroote:
👍  comment added

PR #372, opened 2021-10-29
Remove fixed table of contents scrollbar

dcroote:
This is to improve aesthetics. If the page shrinks vertically to less than the TOC height, a scrollbar will still appear as expected (without shifting text contents). Tested in Chrome, Firefox Edge, Safari.


wkande:
@dcroote Nice catch, looks great.

PR #376, opened 2021-10-31
AN-236 Update contract addresses

bbenligiray:
https://api3dao.atlassian.net/browse/AN-236


wkande (approved):
This look good though I may circle back and change the layout to make the rows more readable.

PR #384, opened 2021-11-02
Document admin CLI docker, add generate-mnemonic command

Siegrift:
https://api3dao.atlassian.net/browse/AN-277

I am not sure where to put these changes - since this is a feature for v0.3 not v0.2.


docs/.vuepress/components/DockerImageVersions.vue
```diff
@@ -21,6 +21,16 @@
           <ExternalLinkImage />
         </a>
       </li>
+
+      <!-- TODO: Uncomment once the airnode admin image is published -->
```
Siegrift, line 25: @wkande This component is common for previous versions as well, so changing this for v0.3 will change this for previous versions of documentation as well.

How are we going to handle this? I think in general, we should be careful about what we treat as "common content"

wkande:
These changes will integrate nicely with the current doc flow. I will make a small adjustment to the DockerImageVersions component to hide future features from `v0.2` where they do not exist.

PR #392, opened 2021-11-05
Recommend version matching

dcroote:
I think it is beneficial to strengthen the language with 'recommend' and state that the minor version and latest strategies are for convenience

### DIFF
--- a/docs/airnode/v0.2/grp-providers/docker/README.md
+++ b/docs/airnode/v0.2/grp-providers/docker/README.md
@@ -25,18 +25,18 @@ system.
 ## DockerHub
 
 Both the Airnode client image and deployer image are available on DockerHub. It
-is not required to download these manually. When you run the Docker commands
-provided these images will be downloaded to Docker automatically.
+is not required to download these manually because they are pulled automatically
+when running the provided Docker commands.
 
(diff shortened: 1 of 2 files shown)


wkande:
@dcroote As it turns out Burak has removed the deployer and client image version `0.2`. I had already started to change the page for Docker (while you where making edits it seems) to use a Vue component which will make it easier to update the version history in a single place. Hopefully there will be a list somewhere in the Airnode repo one day to sync with.

So the changes that you made will not merge. I tried to be conscious of what you wrote but maybe you can check it [here](https://github.com/api3dao/api3-docs/blob/main/docs/.vuepress/components/DockerImageVersions.vue). 

Also I wrote a component on how to delete an AWS based Airnode manually. It is a Vue component as well and it has a pick list for the AWS region in it. In the [docs](https://docs.api3.org/airnode/v0.2/grp-providers/docker/deployer-image.html#manual-removal) and here in the [code](https://github.com/api3dao/api3-docs/blob/main/docs/.vuepress/components/DeleteAirnodeAws.vue). If you have the time give it a read.

dcroote:
@wkande Sounds good, what you wrote captures what I was also thinking 👍

PR #396, opened 2021-11-09
AN-222 Add validator docs

drgy:


wkande (approved):
This look good. I will merge now and place this into v0.3 rather the v0.2. Also the location will most likely change.

PR #401, opened 2021-11-10
api3/deployer renamed api3/airnode-deployer

wkande:


Siegrift (commented):
Ahh, sorry :/ 

This was my fault - I thought that I renamed those in the docs, but I probably renamed just `airnode-client`??? Anyway, thanks for fixing

PR #406, opened 2021-11-12
Document updated reserved params and adapter package

Siegrift:
https://api3dao.atlassian.net/browse/AN-255


wkande (approved):
I see that _type has a major upgrade here. Thanks for the great work on this. Merging now.

PR #408, opened 2021-11-14
Document encoding multiple parameters on chain

Siegrift:
https://api3dao.atlassian.net/browse/AN-125


wkande (approved):
Certainly good to have content. I can learn from this one.

PR #417, opened 2021-11-16
Update the output of testing gateway

Siegrift:
https://api3dao.atlassian.net/browse/AN-355

I am changing the output of the gateway as part of https://github.com/api3dao/airnode/pull/669


wkande (approved):
Appears to be the newer types for a response output? Thanks for getting these updated.

PR #428, opened 2021-11-23
Fix example proposal function signatures

bbenligiray:
See https://docs.soliditylang.org/en/v0.8.10/abi-spec.html#function-selector


wkande (approved):
Guess that was a silly mistake that I missed. I will merge it shortly with a PR I am just finishing up.

PR #433, opened 2021-11-26
Replace relay metadata with relay security schemes

Siegrift:
https://api3dao.atlassian.net/browse/AN-347


Siegrift:
I've also created a https://api3dao.atlassian.net/browse/AN-385 to cleanup and split the api integration file.

wkande (approved):
This helps to understand the changes as a result of removing the _relay-meta_data. I am going to go back through it all after merging to learn this and may have some questions.

PR #441, opened 2021-11-30
Document maximum encoded response length check

Siegrift:
I've only mentioned it in the `request` document and this check doesn't mean much in practice.
As Burak mentioned in the jira issue, saving 1024 bytes string in storage would cost 640K gas and:
1) 640K gas is as of now 1132.112 Ethereum (https://walletinvestor.com/converter/gas/ethereum/640000)
2) 1132 ETH is 5,137,525 USD (https://coinmarketcap.com/converter/eth/usd/)

So we don't need to make a big deal of this check - it's mostly to secure the Airnode from being spammed by large requests.


wkande (approved):
The mention is in a good spot.

PR #455, opened 2021-12-03
Document skip-version-check deployer option

Siegrift:
https://api3dao.atlassian.net/browse/AN-375


wkande (approved):
Simple enough, thanks again.

PR #481, opened 2021-12-15
Explain hardcoded limit for sponsor wallet requests

Siegrift:
https://api3dao.atlassian.net/browse/AN-396


wkande (approved):
👍  Will merge.

PR #484, opened 2021-12-16
Document the EIP-1559 configuration extension to Airnode

aquarat:
Airnode recently added configuration `config.json` options relating to EIP-1559. 
This PR documents those configuration options.


wkande (commented):
So `options` is a field for tx gas fees. Will there be other options in the future? If not then why the name of `options`?

```
"options": {
      "txType": "eip1559",
      "priorityFee": {
        "value": "3.12",
        "unit": "gwei"
      },
      "baseFeeMultiplier": "2"
    },
```

aquarat:
> So `options` is a field for tx gas fees. Will there be other options in the future? If not then why the name of `options`?

My thinking was that perhaps everything chain-specific could eventually be placed under options.
Originally options was named chainOptions, but Burak thought that the chain part was implied by it being part of the chains array.
The other thing I'm thinking is that currently we have hard-coded values for several chain-specific values (in `constants.ts` in `airnode-node`, specifically the `MAXIMUM_SPONSOR_WALLET_REQUESTS` and `BLOCK_COUNT_HISTORY_LIMIT`) and it has been spoken about during the weekly meetings that these things should be exposed as configurable.

PR #490, opened 2021-12-17
SDK now supports versioned templates

wkande:
See the "Use the SDK" section of this doc.

### DIFF
--- a/docs/airnode/v0.4/reference/packages/validator.md
+++ b/docs/airnode/v0.4/reference/packages/validator.md
@@ -71,7 +71,7 @@ console.log(
 );
 // Using a versioned template.
 console.log(
-  validator.validateWithTemplate('exampleSpecs/config@0.4.specs.json', 'config@0.4')
+  validator.validateWithTemplate('exampleSpecs/config.specs.json', 'config@0.4')
 );
 
 # Outputs json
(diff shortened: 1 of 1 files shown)


docs/airnode/v0.4/reference/packages/validator.md
```diff
@@ -60,13 +60,19 @@ api3-validator --template="config" --specs="config.json"
...
 
 ```js
 const validator = require('@api3/airnode-validator');
+// Using the latest template.
 console.log(
   validator.validateWithTemplate('exampleSpecs/config.specs.json', 'config')
 );
+// Using a versioned template.
+console.log(
+  validator.validateWithTemplate('exampleSpecs/config@0.4.specs.json', 'config')
```
drgy: Validator does not care about the file name of the specification, to specify the version that should be validated, it has to be appended to the template name.
```suggestion
  validator.validateWithTemplate('exampleSpecs/config@0.4.specs.json', 'config@0.4')
```

drgy (changes requested):

docs/airnode/v0.4/reference/packages/validator.md
wkande: oops, thanks

wkande:
@drgy This is how your reviewed change committed by GitHub (`#1`). I assume it is OK but the second example (`#2`) is a better example. I just assume that having the version in spec file will just confuse the reader.
```

// #1
console.log(
  validator.validateWithTemplate('exampleSpecs/config@0.4.specs.json', 'config@0.4')
);

/// #2

console.log(
  validator.validateWithTemplate('exampleSpecs/config.specs.json', 'config@0.4')
);

```
Having said that I have talked my self into it so I will make the change and merge. If you see error in my thoughts we can change it. I will make note of this in the JIRA issue as well.

PR #496, opened 2021-12-26
Implement minor beacon language changes

dcroote:


wkande:
@dcroote This would not have merged to the work merged today on the Beacon docs. However the word **tolerance** is a better one than **variance** and has been used. So this PR will be closed. If you get the chance to read over the same doc as it stands now let me know if there are chances for improvement. Also hoping you might be interested in working on the "Why use Beacons?" doc.

PR #502, opened 2021-12-29
Replace erroneous instances of nodeVersion

### LINKED ISSUE #498: nodeVersion on Configuring Airnode looks wrong
In the [nodeVersion](https://docs.api3.org/airnode/v0.3/grp-providers/guides/build-an-airnode/configuring-airnode.html#nodeversion) section of Configuring Airnode it claims the current node version is `1.0.0`, but all the quick deploy examples shows `0.3.1` which looks more correct to be honest.

dcroote:
Fixes GH-498


wkande (approved):
Looks real good, thanks for taking this on.

PR #516, opened 2022-01-02
AN384 proposals

wkande:
So I have made some planned changes for clarity (AN-384) with regards to **Proposal Creation**. BUT since I had screwed up a previous release I do not intent to sink the boat again, at least not alone.

Please review this PR. Below are my points of concern. If the PR changed files view is confusing, please note I have uploaded a [PDF](https://github.com/api3dao/api3-docs/files/7799864/proposals-1.pdf) that shows how the doc would read as this point.

1. Added a **TIP** about USDC decimal precision of 6 places. From an older doc it said to add 6 zeros to the amount you want. So 500,000 would be 500000000000. Is this correct? What if someone wanted 500,000.19 is it 500000190000, not that the docs will speak to it but I am curious. Should this **TIP** be placed in `step #8`. I did not do so as to benefit those that might use the video rather than the text instructions. 

2. Please look at `step #8` in Proposal Creation. Is the json example correct for a value 500,000 USDC? I was concerned that the video shows a huge number and as such maybe I have the wrong understanding of how to write out USDC in the paramters. Please look at the [video](https://youtu.be/XO1iA3wSYMQ) parameters example. 

3. Also please note the paragraph under the json in `step #8`. Does it reflect the proposal call correctly as related to the contract address in `step #5`? 

4. Of course if you could read through the entire doc and express any other concerns or opportunities for improvement I would appreciate it.

[proposals-1.pdf](https://github.com/api3dao/api3-docs/files/7799864/proposals-1.pdf)

### DIFF
--- a/docs/dao-members/contract-architecture/README.md
+++ b/docs/dao-members/contract-architecture/README.md
@@ -20,8 +20,8 @@ for how to deploy a copy of the DAO.
 | DAO Pool         | `0x6dd655f10d4b9e242ae186d9050b68f725c76d76` |
 | Primary Voting   | `0xdb6c812e439ce5c740570578681ea7aadba5170b` |
 | Secondary Voting | `0x1c8058e72e4902b3431ef057e8d9a58a73f26372` |
-| Primary agent    | `0xd9f80bdb37e6bad114d747e60ce6d2aaf26704ae` |
-| Secondary agent  | `0x556ecbb0311d350491ba0ec7e019c354d7723ce0` |
+| Primary Agent    | `0xd9f80bdb37e6bad114d747e60ce6d2aaf26704ae` |
+| Secondary Agent  | `0x556ecbb0311d350491ba0ec7e019c354d7723ce0` |
 | Convenience      | `0x95087266018b9637aff3d76d4e0cad7e52c19636` |
 
 <!-- Add mainnet addresses to this list -->
--- a/docs/dao-members/contract-architecture/voting.md
+++ b/docs/dao-members/contract-architecture/voting.md
@@ -19,9 +19,8 @@ passes.
 
 The API3 DAO has installed two instances of its voting app, primary and
 secondary versions, along with two Aragon Agents that they control. The primary
-app commands a larger treasury and can update all DAO settings, while the
-secondary commands a much smaller treasury and can update some of the DAO
-settings.
+commands a larger treasury and can update all DAO settings, while the secondary
+commands a much smaller treasury and can update some of the DAO settings.
 
 See the
 [Api3Voting.sol](https://github.com/api3dao/api3-dao/tree/main/packages/dao/contracts)
--- a/docs/dao-members/dashboard/proposals.md
+++ b/docs/dao-members/dashboard/proposals.md
@@ -44,21 +44,21 @@ page displays proposals that have been executed or rejected.
 
 1. Navigate to the **Governance** page.
 
-> Here you can browse and create proposals, view the treasury, and delegate your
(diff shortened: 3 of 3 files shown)


dcroote:
I've found a better approach to decimals is:

`(Transfer value to enter) = (Desired number of USDC tokens) * 10**6`

But yes, both of your point 1 maths examples are correct .

bbenligiray:
1. Yes. I think the current placement is fine.

2. Yes.

3. I found the wording to be a bit packed, here's a more verbose version
> In the example above, the respective Agent (primary or secondary) would be calling the USDC contract (`0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`) to transfer 500,000 USDC to `0xF4EB52Cf9D31a...d1663d78ddDEE9`. Note that since `transfer(address,uint256)` transfers funds from the sender to the specified address, the USDC is asked to be supplied from the Agent's balance.

bbenligiray:
I went through the whole document and I think the proposal execution section has a minor error. When >50% of all voting power votes yes on a proposal, the last vote that carries the quorum over the threshold also executes the proposal. In other words, there is no case where a proposal has 50%+ quorum and is waiting to be executed. So it should read

> A proposal is ready for execution if:
> 1. The proposal hasn't already been executed, and
> 2. the proposal's voting period has ended, and
> 3. the total "yes" vote exceeds the "no" vote, and
> 4. (for Secondary type proposals) at least 15% of all voting power has voted "yes" on the proposal.
>
> Primary type proposals require 50% of all voting power to have voted "yes" on the proposal. Both primary and secondary type proposals execute immediately once 50% of all voting power has voted "yes" on them.

bbenligiray:
Another thing, note that the concepts "Agent", "apps" and "shares" are never mentioned on the DAO dashboard. We abstracted these away because they are not useful for the user and only create confusion. We have a primary/secondary treasury (not 
Agent apps). We have primary and secondary proposal types (not proposals created on primary and secondary Voting apps). When you stake 1% of the current staked amount, you will get 1% of the total voting power (you don't know or care about how many shares you have or that there is even such a concept).

wkande:
@bbenligiray So your comment seems to imply the docs should not use "Agent", "apps" and "shares" much like the DAO Dashboard. Having said that then the example you provided above that describes the execution of ` transfer(address,uint256)` would need to be re-written.

> In the example above, a successful proposal would automatically execute the USDC contract (`0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`) to transfer 500,000 USDC to `0xF4EB52Cf9D31a...d1663d78ddDEE9`. Note that since `transfer(address,uint256)` transfers funds from the appropriate treasury to the specified address, the USDC is asked to be supplied from the treasury.

Siegrift:
I've read the whole DAO members section now and I've implemented some changes in this commit: https://github.com/api3dao/api3-docs/commit/2f0e6fcb060ff8eaeb7fae815455ae2cf35b5b5e

> I found the wording to be a bit packed, here's a more verbose version

I preferred the original version as the new one sounds too low level (and we try to avoid that)

bbenligiray:
> So your comment seems to imply the docs should not use "Agent", "apps" and "shares" much like the DAO Dashboard.

Ideally, as much as possible. It is not a very strict requirement.

> new one sounds too low level

I think it is required here because it may not be obvious to the user that when they make a proposal, it is the agent that will make the transaction and not them. In general I like writing detailed docs and always received good feedback about it. The user either already knows that stuff and can skim over it or they don't know and appreciate the information.

PR #517, opened 2022-01-04
Move "testable" property from OIS to triggers.rrp section

Siegrift:
https://api3dao.atlassian.net/browse/AN-332


wkande (approved):
Yes everything looks good on the testable move.

PR #518, opened 2022-01-04
Correct the wrong OIS version

Siegrift:
https://api3dao.atlassian.net/browse/AN-456


Siegrift:
The convertor documentation is still broken though as I described in https://api3workspace.slack.com/archives/C02AYRX8D89/p1641315632468100

For example, the latest convertor (v3.1) should be called like this:
```
npx --package=@api3/airnode-validator -c "airnode-convertor --from=ois@0.2 --to=config@0.2 --specs=exampleSpecs/ois.specs.json"
```
and we are missing a lot of conversion pairs.

wkande (approved):
Thanks

PR #533, opened 2022-01-14
Add derive-airnode-address command, change "airnode-address" parameter

Siegrift:
concepts/airnode.html#airnodeaddress


wkande (approved):
This looks accurate to me.

PR #534, opened 2022-01-14
Clarify admin CLI commands for withdrawals

Siegrift:
Byproduct of the https://api3dao.atlassian.net/browse/AN-410

@wkande This might be worth clarifiying in previous versions. WDYT?


wkande (approved):
Good clarity

PR #535, opened 2022-01-14
Make the protocol ID start from 1

Siegrift:
https://api3dao.atlassian.net/browse/AN-453

### DIFF
--- a/docs/airnode/v0.4/concepts/sponsor.md
+++ b/docs/airnode/v0.4/concepts/sponsor.md
@@ -56,7 +56,7 @@ Note that a sponsor could use multiple addresses from multiple wallets. Below
 are some example reasons why one would want to have multiple `sponsorAddress`
 identities on-chain:
 
-- To keep separate `sponsor-wallets` for two separate use-cases for easier
+- To keep separate `sponsorWallets` for two separate use-cases for easier
   accounting.
 - To duplicate transaction queues for a single use-case and increase response
   throughput.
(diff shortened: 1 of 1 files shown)


docs/airnode/v0.4/concepts/sponsor.md
```diff
@@ -82,25 +82,17 @@ from different sponsor wallets are performed in parallel.
...
-wallet non-hardened derivation path goes up to 2^31. This requires the division
-of these 160 bits into six 31 bit-long chunks, therefore derivation path for a
-sponsor wallet of a requester would be:
-
-<!-- TODO: The Derivation Path shown below will change in v0.4. Watch the
-JIRA issue AN-453. -->
+An Ethereum address is 20 bytes-long, which makes it 160 bits. Each index in the
+HD wallet non-hardened derivation path goes up to 2^31. This requires the
+division of these 160 bits into six 31 bit-long chunks, therefore derivation
+path for a sponsor wallet of a requester would be:
```
wkande: Are you sure that we should remove lines 89-97. I think Burak put these here but maybe not. If not he did so somewhere else in the docs. Do you think it is overkill?

Siegrift: I removed it since PSP is now postponed so reading something like "pub-sub" is confusing. Since a thing that doesn't and will not exist for some time.

acenolaza: Should we still mention somehow that the `/1/` in the derivation path means that is a sponsorWallet specific to RRP, 2 for PSP and that for Airkeeper the `protocolId` is 12345?

Siegrift: I don't think this is relevant to the end user and it sounds too technical to me.

wkande: I need to research this more. To some extent you are right as it is overly technical for readers external to API3. But the docs also try to support internal API3 members like Cameron's team.

Siegrift: Yeah, but low level technical details intended for developers are to be documented in the README of the respective package. What cameron (and also users) need is the information that:
"A sponsor wallet derived for a particular sponsor depends on the Airnode protocol (rrp, beacon or later psp). This means that a sponsor wallet address derived for a particular sponsor will be different for RRP or beacons. This is intentional and allows a particular sponsor to only fund a particular protocol (e.g. only beacons)".

wkande: Perhaps that is the solution. Provide a link to the README that contains the technical info. We would need to make sure that README is up to date with what is stated in the docs. Need to think more on this one.

bbenligiray: A lot of the users that are not Airnode developers will want to know how this works and they shouldn't have to refer to the repo README or the code for that. The reasoning behind the convention is helpful in understanding it and the fact that there are multiple protocols is important for that, otherwise there is that `/1/` in the path that doesn't really do anything and that's also confusing. Overall, more information is better and if it sounds too high IQ then maybe that's a good thing.

wkande: I will pull this branch and work on it to preserve the High IQ stuff and changes that the PR introduces that have correction and new add-value content. If ya'll trust me to do so?

bbenligiray: Maybe say something along the lines of "Don't worry if you don't understand this part, just use the addresses the admin derives for you"

PR #543, opened 2022-01-19
AN-497: Unify "scheme/schema" spelling to "scheme"

martinkolenic:


wkande:
@kolenic-martin Thanks for doing this. Time is a crunch for us all right now.

PR #545, opened 2022-01-19
BEC98 contract addresses and beaconIds

wkande:
This PR contains the latest `beaconIds` from @aquarat  from his `documentation_beacons_lite-json file`.

It also uses latest `RrpBeaconServer` contract addresses (`v0.3.1`) from @Ashar2shahid  using the [beacon-setup-guide repo](https://github.com/api3dao/beacon-setup-guide/tree/main/deployments/0.3.1).

I tested them using ethers.js (direct to the `RrpBeconServer` contract) and they all appear to work.


Ashar2shahid:
Ropsten seems to be missing from this list but other than that this looks right

PR #557, opened 2022-01-23
Add string32 and bool options to Airnode ABI

Siegrift:
https://api3dao.atlassian.net/browse/AN-459


wkande (approved):
Everything seems to match task JIRA AN-459 best I can tell. Someone better than me probably could do a better review but in the end you seem to get it right every time so let's merge this.

Siegrift:
We can maybe wait till the PR in Airnode is merged. We iterated multiple times already on the design of this, but I am hopeful this is gonna be the final version.

dcroote:
@Siegrift - as an aside it looks like the branch deployment workflow didn't run because the workflow file isn't present- did you pull from `main` before branching?

Siegrift:
Not sure, but we should setup "require branches to be up to date" in github settings, so that you have to rebase to latest main.

PR #563, opened 2022-01-26
Add warning against simultaneous deployments

amarthadan:
Related to [AN-499](https://api3dao.atlassian.net/browse/AN-499)

I know that GCP technically doesn't need the warning but I would rather have it everywhere.

### DIFF
--- a/docs/airnode/v0.4/reference/packages/deployer.md
+++ b/docs/airnode/v0.4/reference/packages/deployer.md
@@ -141,6 +141,14 @@ When completed the `deploy` command creates a receipt using the path and name
 from the `--receipt` argument. The receipt contains metadata about the
 deployment and can be used to remove the Airnode.
 
+::: warning Simultaneous deployments
+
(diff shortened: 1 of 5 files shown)


docs/airnode/v0.4/reference/packages/deployer.md
```diff
@@ -141,6 +141,14 @@ When completed the `deploy` command creates a receipt using the path and name
 from the `--receipt` argument. The receipt contains metadata about the
 deployment and can be used to remove the Airnode.
 
+::: warning Simultaneous deployments
+
+Please, avoid running multiple deployment commands simultaneously. Doing so
```
Siegrift, line 146: You could mention that it's still possible to remove manually by going into AWS.

wkande (approved):
I will go ahead and merge this. However I have an idea that I might follow-up on later. That is to create a very simple Vue component with the content here and also to make reference to what @Siegrift  has suggested. This way the content will be in one central component easy to change whenever needed.

PR #575, opened 2022-02-01
Proofread: Improve airnode's main page readability and streamline syntax

martinkolenic:
The text on the main page of airnode documentation wasn't bad per se (actually, it was a nice read!); however, having in mind that most devs who will be reading it might not necessarily come from an Anglophone environment, I made some changes and simplifications in terms of lexis and syntax.


martinkolenic:
@wkande it seems I have broken the formatting; sorry about that. Could you perhaps help me out? I've looked at the file but cannot find the problematic spot(s).

amarthadan:
> @wkande it seems I have broken the formatting; sorry about that. Could you perhaps help me out? I've looked at the file but cannot find the problematic spot(s).

@kolenic-martin You can use `yarn format:check` to see what's wrong and `yarn format` to fix it :slightly_smiling_face:

PR #576, opened 2022-02-01
Proofread: airnode main page edits

martinkolenic:
I've recreated the pull request as the previous one failed tests. I reran Prettier and everything seems to pass on my side. I'm hoping to pass the tests server-side as well.

PR #577, opened 2022-02-01
Skip etherscan in CI link check

dcroote:
Etherscan has been consistently giving 403 errors on CI e.g. #576


wkande:
@dcroote @kolenic-martin 
Yah that bad boy has been annoying. It has been an issue in the link validator library as well. I have plans to implement
a retry for all external links in the link validator as I believe a second call most likely will succeed. **[Issue 578](https://github.com/api3dao/api3-docs/issues/578)**

PR #590, opened 2022-02-08
Document the services CLI tool for creating beacon reader project

Siegrift:
https://api3dao.atlassian.net/browse/BEC-142

### DIFF
--- a/docs/beacon/v0.1/functions/README.md
+++ b/docs/beacon/v0.1/functions/README.md
@@ -27,12 +27,37 @@ individually or combined to build decentralized data feeds or
 - [requestBeaconUpdate()](./request-beacon-update.md) - Triggers an update to a
   Beacon's value.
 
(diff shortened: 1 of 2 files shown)


wkande (approved):
This looks good. What I want to do is merge and make a few minor edits to the content for grammar. Also I thought that since you have the same content in two different places maybe I would create a component to display the content. However I am not sure it may be needed because the `hackathon.md` page goes away in April. But we may want to show the content in other places in the future. WDYT?

Siegrift:
> But we may want to show the content in other places in the future.

I think we can create the component for it when the time comes. I am not sure we will need to duplicate it. The services package will grow, so it will likely desrve it's own page where this CLI can be documented and other places can just link to it (similarly how we do it in Airnode docs)

vponline (approved):
Looks good to me as well

PR #595, opened 2022-02-08
v0.4 Airnode doc set release

wkande:
Make the Airnode document set for `v0.4` the latest/current. The Beacon and OIS versions remain as they are.


wkande:
@dcroote There are many broken links in this PR until @Burak creates the `v0.4` tag in the `airnode` repo.

Of course it is strange that the lychee link checker did not pick them up. None-the-less the link validator library did so we are ok.

dcroote:
@wkande - yes that's expected [because of this line](https://github.com/api3dao/api3-docs/blob/a4dcb70343ca6ecc37bfaeb85ecd9717f82c3d99/.github/workflows/check-links-imgs.yaml#L40), which should be either deleted or kept, but with `v0.4` replaced with `v0.5`

PR #605, opened 2022-02-10
Beacon docs review

bbenligiray:
I'm not finished yet because the commands don't work for me. Still can be merged as a first pass.


wkande (approved):
I will merge this but will not delete the branch.

PR #606, opened 2022-02-10
Replace "npx" with "npm exec" for services CLI

Siegrift:
The `npx` command seems to fail because of the github dependencies that are used in services.
I am saying "seems", because I am not sure, but running `npx --loglevel verbose --package @api3/services --call "create-beacon-reader-app --help"` fails after downloading the utility contracts package.

Also, there are some issues on the NPM repo mentioning problems with github dependencies:
https://github.com/npm/cli/issues/4137
https://github.com/npm/cli/issues/4003

I don't think many people use them in production, so that's why I expect I didn't find the exact issue. Gladly, I empirically found out that replacing `npx` with `npm exec` works. Also `npm i @api3/services` works. Also, after doing `npm i` the `npx` starts to work (likely because it uses the cached package downloaded by npm - also this was probably a reason why the package worked for people testing https://api3dao.atlassian.net/browse/BEC-138).

### DIFF
--- a/docs/beacon/v0.1/functions/README.md
+++ b/docs/beacon/v0.1/functions/README.md
@@ -36,7 +36,7 @@ generates the minimal project files that will get you started with building your
 application based on beacons. Simply run:
 
 ```
-npx --package @api3/services --call create-beacon-reader-app
+npm exec --package @api3/services --call create-beacon-reader-app
 ```
 
 The CLI tool will ask you for path in which to initialize the project and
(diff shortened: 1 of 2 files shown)


Siegrift:
@wkande please wait with merging until @bbenligiray confirms this hack works.

wkande:
Will not merge until you give the OK.

bbenligiray:
When I run `npm exec --package @api3/services --call create-beacon-reader-app` I get `sh: 1: create-beacon-reader-app: not found`

`npx --loglevel verbose --package @api3/services --call create-beacon-reader-app` also says `sh: 1: create-beacon-reader-app: not found`

dcroote:
Different npm versions may be behaving differently? I reproduced the `npx` error, realized I was using npm 6.x and it doesn't have `npm exec`, then upgraded npm with `npm install npm@latest -g` to `8.5.0`. Next `npx` asked to install `@api-services`, I confirmed yes, and it appeared to work, except didn't because the next time I ran the same command it asked the same thing again. However, `npm exec` did work, following which `npx` commands started working.

bbenligiray:
I'm using npm v8.1.2 and this is exactly what happened to me
> Next npx asked to install @api-services, I confirmed yes, and it appeared to work, except didn't because the next time I ran the same command it asked the same thing again. However, npm exec did work, following which npx commands started working.

except I had to specify the @api3/services version even though my cache is empty.

bbenligiray (approved):
Reportedly helps

PR #613, opened 2022-02-11
Document `maxConcurrency` option for HTTP gateway

amarthadan:
[Issue AN-522](https://api3dao.atlassian.net/browse/AN-522)


wkande (approved):
This looks really good, thanks.

PR #620, opened 2022-02-13
Remove references to airnode-client-examples repo

dcroote:
As it no longer exists: https://github.com/api3dao/airnode-client-examples/


bbenligiray:
Thanks, I made that repo private because I think it will probably do more harm than good

PR #626, opened 2022-02-16
Document signed relayed triggers section

Siegrift:
https://api3dao.atlassian.net/browse/AN-352

### DIFF
--- a/docs/airnode/v0.5/grp-providers/guides/build-an-airnode/configuring-airnode.md
+++ b/docs/airnode/v0.5/grp-providers/guides/build-an-airnode/configuring-airnode.md
@@ -361,8 +361,9 @@ for the gateway.
 
 List the endpoints that you want to serve with the request–response protocol
 (RRP) under `triggers.rrp`. List the endpoints that you want to serve with the
-HTTP gateway under `triggers.http`. In most cases, you would create a trigger
-for each endpoint in your OIS object.
+HTTP gateway under `triggers.http`. List the endpoints which can be used to get
+the signed relayed data in `triggers.httpSignedRelayed`. In most cases, you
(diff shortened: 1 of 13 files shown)


docs/airnode/v0.5/grp-providers/guides/build-an-airnode/configuring-airnode.md
```diff
@@ -396,6 +397,39 @@ protocol. Only endpoints listed here will be served through the RRP protocol
...
+
+- endpointId: Add an `endpointId` to the trigger which is the ID that a
+  requester will use for on-chain requests to reference a specific trigger. Use
+  the admin CLI command
+  [derive-endpoint-id](../../../reference/packages/admin-cli.md#derive-endpoint-id)
+  to derive endpoint IDs using the `oisTitle` and `endpointName`.
+
+#### httpSignedRelayed
+
+<!-- TODO: This is a difficult concept so it should be documented in a separate section. -->
```
wkande, line 433: Not sure yet where to do this but I will make a GitHub issue for it and we can go over it later. It does appear the concept is harder to grasp.

amarthadan:
I added my changes, at least those I could.

I didn't want to create additional sections or move stuff regarding the second gateway because I don't really know what the best place would be, probably the same place we decide to put @Siegrift's explanation of the `httpSignedRelayed` section.

Basically, we're adding another HTTP gateway called signed relayed HTTP gateway. The new one will require an ID and relayer's address (already described) and will return a value with the signature. The old HTTP gateway doesn't sign the value as it's not necessary and computationally heavy.

So, I'm not sure if we want a separate section for the new gateway or just expand the current HTTP gateway section somehow.
@wkande Once you set up a structure we're happy to fill in the blanks.

PR #635, opened 2022-02-18
AN149-fix: Remove `value` from CLI arguments

vponline:
Removing `value` as an admin CLI argument as it is not needed in the implementation.


wkande (approved):
Thanks for keeping things up-to-date and proofing as well.

PR #637, opened 2022-02-20
Add descriptions to landing page buttons

dcroote:
Adds images and descriptive text on the landing page for users unfamiliar with the project.

### DIFF
--- a/docs/.vuepress/components/DocumentSetButtons.vue
+++ b/docs/.vuepress/components/DocumentSetButtons.vue
@@ -8,17 +8,19 @@ it will always default to the startPath in config.json.
 <template>
   <div class="landing-container">
     <li>
-      <img src="/img/01-Airnode-State=Active@2x.png" />
-      <router-link class="nav-link" :to="latestVersion">Airnode</router-link>
+      <router-link :to="latestVersion">
+        <img src="/img/01-Airnode-State=Active@2x.png" />
+        Airnode</router-link
+      >
       <p>
         First-party serverless oracle node that solves the Web3 API connectivity
         problem
       </p>
     </li>
     <li>
-      <img src="/img/Beacons-active.png" />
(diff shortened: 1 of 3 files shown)


wkande:
Changing the DocumentSetButtons Vue component should be simple enough. When adding styles to the file use longer names like doc-set-btns-text. I discovered the css can have impact on other components if they use the same name.
```
 doc-set-btns-text{
   ...
}
```

We should probably change the existing style from `btn` to `doc-set-btn` as well.

dcroote:
@wkande - have a look now, I've added images and improved the text descriptions, being mindful of both desktop and mobile

wkande:
@dcroote - Can the router-link wrap the image and description while keeping the heading font green and the paragraph font black? This makes it easier for the reader to select a document set.

Also can we center the button and heading above the paragraph. It might look better though that remains to be seen.

dcroote:
@wkande  - I centered the doc sets, including the descriptive text (it looked funny at some screen widths not being centered along with the image and header link).

I also made the images clickable. I didn't make the descriptive text itself clickable as I think that would be a bit unexpected and probably lead to accidental clicks, especially when scrolling on mobile.

To see the updated site you made have to do a hard refresh to clear the cache.

wkande:
Now I'm just getting picky. The icons seem so black and gloomy. What do you think about adding `opacity: .7` or so to lighten them up? Also the "OIS" Link is getting pushed down by its non-standard fixed icon. Can the "OIS" be moved up: `margin-top:-7px` or so? That way all the doc set names line up on the desktop browser. And maybe some periods in the description. Seems funny without them when the blocks below have them.

dcroote:
All good suggestions! I avoided playing with the margin for the OIS image by instead setting all img heights to be fixed rather than width

wkande (approved):
Look clean

PR #649, opened 2022-02-25
Update deployer CLI examples to match airnode

vponline:
This PR is about the changes made here: [#869](https://github.com/api3dao/airnode/pull/869/files#diff-0163e28f7c9809df7e745f7f5ab8376f0cdfa5635762d13c3107f5047cc967bcR79)

### DIFF
--- a/docs/airnode/v0.5/reference/packages/deployer.md
+++ b/docs/airnode/v0.5/reference/packages/deployer.md
@@ -157,7 +157,7 @@ Options:
   -r, --receipt                          Output path for receipt file          [string] [default: "output/receipt.json"]
 
 # Example
-deployer deploy --config myConfig/config.json --secrets myConfig/secrets.env -r myOutput/receipt.json
+deployer.ts deploy -c pathTo/config.json -s pathTo/secrets.env -r myOutput/receipt.json
 ```
 
 ### remove
--- a/docs/airnode/v0.5/reference/packages/deployer.md
+++ b/docs/airnode/v0.5/reference/packages/deployer.md
@@ -157,7 +157,7 @@ Options:
   -r, --receipt                          Output path for receipt file          [string] [default: "output/receipt.json"]
 
 # Example
-deployer.ts deploy -c pathTo/config.json -s pathTo/secrets.env -r myOutput/receipt.json
+airnode-deployer deploy -c pathTo/config.json -s pathTo/secrets.env -r myOutput/receipt.json
 ```
 
 ### remove
@@ -190,6 +190,6 @@ Options:
   -p, --project-id             Project ID (GCP only)                                                            [string]
 
 # Examples
-deployer remove --receipt myOutput/receipt.json
-deployer remove --airnode-address-short abd9eaa --stage dev --cloud-provider aws --region us-east-1
+airnode-deployer remove --receipt myOutput/receipt.json
+airnode-deployer remove --airnode-address-short abd9eaa --stage dev --cloud-provider aws --region us-east-1
 ```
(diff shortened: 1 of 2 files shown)


vponline:
Is there a need to add the output of `npx @api3/airnode-admin --help` to the admin-cli docs?

docs/airnode/v0.5/reference/packages/deployer.md
```diff
@@ -157,7 +157,7 @@ Options:
   -r, --receipt                          Output path for receipt file          [string] [default: "output/receipt.json"]
 
 # Example
-deployer deploy --config myConfig/config.json --secrets myConfig/secrets.env -r myOutput/receipt.json
+deployer.ts deploy -c pathTo/config.json -s pathTo/secrets.env -r myOutput/receipt.json
```
Siegrift: That is not true. The binary will be called `airnode-deployer` https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/package.json#L7

Also, this section is outdated: https://docs.api3.org/airnode/v0.5/reference/packages/deployer.html#global-package (and probably older version are as well) (cc: @wkande )

wkande: @Siegrift Which older version are outdated? `v0.4`? I am not sure as to when the package name changed.

Siegrift: It should change for v0.3 at least, you can see this when you execute: `npx @api3/airnode-deployer@0.3.1 --help` 

(I'd not care about the other versions, since they are not used by anyone)

docs/airnode/v0.5/reference/packages/deployer.md
```diff
@@ -56,7 +56,7 @@ yarn global add @api3/airnode-deployer
 npm install @api3/airnode-deployer -g
 
 # Executing the deployer.
-api3-deployer deploy --config myConfig/config.json --secrets myConfig/secrets.env -r myOutput/receipt.json
+airnode-deployer deploy --config myConfig/config.json --secrets myConfig/secrets.env -r myOutput/receipt.json
```
Siegrift: Nit: It would be nice if this was consistent with the lines below.

docs/airnode/v0.5/reference/packages/deployer.md
Siegrift: Just to clarify, the package name didn't change https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/package.json#L2 only the name of the binary https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/package.json#L7

But it seems to me we will not care about the older versions based on the discussion here: https://api3workspace.slack.com/archives/C02AYRX8D89/p1645787741584089

docs/airnode/v0.5/reference/packages/deployer.md
vponline: Thanks! Fixed these.

docs/airnode/v0.5/reference/packages/deployer.md
vponline: Just to confirm that then we are not changing `0.3` and `0.4` `deployer.md`s to be consistent with these updates?

Siegrift: We should change it for 0.4 (since that is the current supported version) and ignore 0.3.
Also, it should change for 0.5 (which is the next one).

vponline: Ok great, I've fixed the deployer commands for v0.4 now as well so this should be good to go now.

wkande (approved):
@vponline @Siegrift - Thanks for keeping things up-to-date guys. This type of interaction with the docs really helps.

PR #650, opened 2022-02-25
Proofread

martinkolenic:
This is the first half of the grp-developers section I worked on today and yesterday. I made changes to v0.4 and v0.5 sections (they should be the same atm), mostly stylistics- and syntax-related.


Siegrift (commented):
🥇 LGTM, I think those touches are easier to read :)

wkande (approved):
Thanks for the changes. The soft touches you add are nice.

PR #651, opened 2022-02-26
AN568 redirect fuzzy terms

wkande:
Dynamic browser base redirects. The redirect plugin is retired. Redirects now happen within the client. Additionally, support has been added to `Sidebar.vue` to honor the anchor (hash) in the startup URL for the SPA.

So Aaron started this change in PR: 636. I have updated it to honor the browser history (back button) and added a trap for the activation of anchors when the SPA loads. These are implemented separately.

- [`enhanceApp.js`](https://github.com/api3dao/api3-docs/pull/651/files#diff-1e710817ef664ac4cec94f942fde94da9507c661f20cfb620d550ccd13f346c6): redirects
- [`Sidebar.vie`](https://github.com/api3dao/api3-docs/blob/main/docs/.vuepress/components/Sidebar.vue#L33-L43): anchors (previous PR)

A few files have been archived (`/archived`) by the PR, they can be ignored.

Examples (partial):
```
{
    from: `/airnode-starter`,
    to: `/airnode/pre-alpha/tutorials/airnode-starter.html`,
    exact:true
    // comment: 'legacy',
  },
  // fuzzy
  { from: `/pre-alpha`, to: `/airnode/pre-alpha/`, fuzzy: true }, // comment: 'legacy'
  { from: `/latest/members`,to: `/dao-members/`, fuzzy: true },
  { from: `/latest`, to: `/airnode/${airnodeLatest}/`, fuzzy: true },
  { from: `/airnode`, to: `/airnode/${airnodeLatest}/`, fuzzy: true},
  { from: `/airnode/latest`, to: `/airnode/${airnodeLatest}/`, fuzzy: true },
  { from: `/airnode/next`, to: `/airnode/${airnodeNext}/`, fuzzy: true },
  { from: `/next`, to: `/airnode/${airnodeNext}/`, fuzzy: true },
  { from: `/beacons`, to: `/beacon/${beaconLatest}/`, fuzzy: true },
  { from: `/beacons/latest`, to: `/beacon/${beaconLatest}/`, fuzzy: true },
  { from: `/beacon`, to: `/beacon/${beaconLatest}/`, fuzzy: true },
  { from: `/beacon/latest`, to: `/beacon/${beaconLatest}/`, fuzzy: true },
  { from: `/ois/latest`, to: `/ois/${oisLatest}/`, fuzzy: true },
  { from: `/ois`, to: `/ois/${oisLatest}/`, fuzzy: true },
...
```

One think to check is whether it is needed to have the two checks, exact and fuzzy. Maybe fuzzy is enough. I have updated the fuzzy check to allow a redirect only if the `to.path` is not equal to the redirect `from` key. Probably not and leaving things alone is probably the safer bet.

```
// exact
const exactRedirect = redirects
    .filter((item) => item.exact)
    .find((item) => to.path === item.from && item.to);

// fuzzy
const fuzzyRedirect = redirects
    .filter((item) => item.fuzzy)
    .find((item) => to.path.indexOf(item.from) === 0 && to.path != item.to);
```


dcroote:
In the Firebase preview deployment, when I use a `latest` redirect, I briefly get a `404` page before it continues on to the correct page. Notwithstanding, the link anchor redirects appear to work after continuing past the `404`. 

Try here: https://api3-docs--an568-redirect-latest-m94yx89n.web.app/airnode/latest/grp-providers/airnode/design-philosophy.html#requirements

aquarat:
Hey Warren, I'll check properly on Monday, but it sounds great (!), well done.
Firebase can be configured to redirect invalid URLs temporarily to any page instead of a 404. I used the index page in my (outdated) PR (as the index page will load the SPA/Router):
https://github.com/api3dao/api3-docs/blob/60c81667143c6196a56f6ba6100e8c4cb847a077/firebase.json

wkande:
I see that. I do wonder if GitHub will present the same behavior. Not the end of the world but I might try `router.push` rather than` router.redirect` to see if it goes away.

wkande:
This might also happen on GitHub pages. The 404 is needed for other URLs that the user might type in so on GitHub we cannot replace it with `index.html`. The 404 page appears before the function in `enhanceApp.js` fires I believe. Not sure if we can tap into the router stack before that happens. Will need to check.

wkande:
@aquarat @dcroote @kolenic-martin  I added two new commits to this PR. 

- The first [**here**](https://github.com/api3dao/api3-docs/pull/651/commits/a41ac05fb8f7c4ffe7b6f4e09267300a34f81eba), adds a check to see if the requested route is in the list of known routes. If it is then checking the redirects is senseless and thus avoided.

```
const route = router.options.routes.find((obj) => obj.path == to.path);
```

- The other [**here**](https://github.com/api3dao/api3-docs/pull/651/commits/1e9c00d631a667a3868af097817c333f39cc1a24), moves Vuepress's default theme 404 component to the custom components folder (named: `Replacement404.vue`). Now I get to mess with it and its executing script. If the blinking 404 shows up on GitHub pages (I'm sure it will), I will be able to blank it out for things that are redirected such as routes that contain say `/latest`. Having options is nice.

archived/redirects-hold.js
```diff
@@ -0,0 +1,114 @@
...
+    exact: true, // comment: 'legacy',
+  },
+  {
+    from: `/d/response-parameters`,
+    to: `/airnode/${airnodeLatest}/grp-developers/call-an-airnode.html#response-parameters`,
+    exact: true, // comment: 'legacy',
+  },
+  {
+    from: `/r/reserved-parameters`,
+    to: `/airnode/${airnodeLatest}/reference/specifications/reserved-parameters.html`,
```
aquarat: It's great that you've made these all template based, it'll make future versioning much easier to maintain.

archived/enhanceApp copy.js
```diff
@@ -0,0 +1,80 @@
...
+
+  const exactRedirect = redirects
+    .filter((item) => item.exact)
+    .find((item) => to.path === item.from && item.to);
+
+  if (exactRedirect) {
+    return { ...to, path: exactRedirect.to, replace: true };
+  }
+
+  //let fuzzyRedirect = undefined;
```
aquarat: Nit: you can remove this now as you've got it working nicely? :)

archived/redirects-hold.js
```diff
@@ -0,0 +1,114 @@
...
+    fuzzy: true,
+  },
+  {
+    from: [
+      '/beacons/',
+      '/beacons',
+      '/beacons/latest/',
+      '/beacons/latest',
+      '/beacon/',
+      '/beacon',
```
aquarat, line 83: Some of these overlap with the destination address... will this be a problem? It's the reason I sorted the other redirects (long paths first so they match first).

aquarat (changes requested):
Fantastic work @wkande 
I consider these docs to be extremely important as they're a primary entry point for the public. A bit of love like this will go a long way.
I had a concern about some of the redirect routes overlapping, but it doesn't seem to impact functionality (yay).

archived/enhanceApp copy.js
wkande: No problem, this is actually a file moved to the `/archived` directory. It will be deleted.

archived/redirects-hold.js
wkande: In this file I tried an array based approach to manage the from routes. I ended up abandoning the approach and moved the file to the archive directory. It will be deleted.

PR #654, opened 2022-03-03
Rename signed HTTP gateway

amarthadan:


wkande (approved):
Looks like you go them all. Later can we DM or talk about a simple use case for this.

PR #675, opened 2022-03-17
Add cloud resource descriptions

amarthadan:
[AN-592](https://api3dao.atlassian.net/browse/AN-592)

@wkande Feel free to change the wording or anything. One could argue that these are not all the resources (e.g. HTTP gateway is composed of multiple pieces - gateway, deployment, stage, usage plan, etc.) but I think this should be good enough.


wkande (approved):
Looks good thanks.

PR #676, opened 2022-03-18
GCP project setup changes

wkande:
Some changes to the instructions when creating a new GCP project.

### DIFF
--- a/docs/airnode/v0.5/grp-providers/tutorial/quick-deploy-gcp/README.md
+++ b/docs/airnode/v0.5/grp-providers/tutorial/quick-deploy-gcp/README.md
@@ -114,25 +114,27 @@ Add values for each of the these fields.
 
 ### GCP Project Setup & Credentials
 
-First, you need to
-[create a GCP project](https://cloud.google.com/resource-manager/docs/creating-managing-projects)
-under which will the Airnode be deployed. Once the project is created, insert
-its project ID into your
-[config.json](../../guides/build-an-airnode/configuring-airnode.md#cloudprovider).
+1. First, you need to
+   [create a GCP project](https://cloud.google.com/resource-manager/docs/creating-managing-projects)
+   under which will the Airnode be deployed. Once the project is created, add
+   the project ID to the [secrets.env](./secrets-env.md) file.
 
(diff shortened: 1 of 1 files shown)


docs/airnode/v0.5/grp-providers/tutorial/quick-deploy-gcp/README.md
```diff
@@ -114,25 +114,27 @@ Add values for each of the these fields.
...
+   under which will the Airnode be deployed. Once the project is created, add
+   the project ID to the [secrets.env](./secrets-env.md) file.
+
+2. In order for Airnode to deploy successfully, you need to enable the
+   [App Engine Admin API](https://console.cloud.google.com/apis/library/appengine.googleapis.com)
+   specifically for the project. After enabling it, wait a few minutes before
+   deploying the Airnode for this change to take effect.
+
+3. Enable the
+   [API Gateway](https://console.cloud.google.com/marketplace/product/google/apigateway.googleapis.com?returnUrl=%2Fapi-gateway%2Fapi&project=zzz).
```
amarthadan, line 128: Enabling the HTTP gateway manually should not be needed. Did you have to do that? Also, the link won't work because it contains the project name (at least I presume that `zzz` is the name from the URL).

wkande: @amarthadan 
> Enabling the HTTP gateway manually should not be needed. Did you have to do that?

Yes, but I can try it again. This was an earlier error I got. But for now I cannot get to it because of the storage error permissions.


---

>Also, the link won't work because it contains the project name (at least I presume that zzz is the name from the URL).

I think I forgot to add a field for the user to add their project ID. But if manual enabling the GCP Gateway is not needed I can remove this step.

```
(https://console.cloud.google.com/marketplace/product/google/apigateway.googleapis.com?returnUrl=%2Fapi-gateway%2Fapi&project=myProjectID).
```

amarthadan: I've asked @kolenic-martin to test it, we'll get back to you.

PR #680, opened 2022-03-22
Ignore rinkebyfaucet in CI

dcroote:


wkande (approved):
These sites we are omitting seem to be an issue. The link validator seems to see them more often than not so for the most part the links seem valid. I know some are doing a rewrite.

PR #684, opened 2022-03-25
Fix signed data Airnode example

Siegrift:
https://api3dao.atlassian.net/browse/AN-558


wkande (approved):
Thank for the clean-up.

PR #691, opened 2022-03-31
Add information on enabling billing for the GCP project

martinkolenic:
Since we've encountered issues on projects that did not have billing enabled, I thought it might be beneficial for the user to know this needs to be done to test GCP deployment (although there will be no charges). Feel free to change the wording.


wkande (changes requested):
This also needs to be done for `v0.6`. Can you add it otherwise I can do it later.

martinkolenic:
> This also needs to be done for `v0.6`. Can you add it otherwise I can do it later.

Will do.

PR #704, opened 2022-04-14
Fix encode value for MyBytes32

wkande:
Incorrect value for `MyBytes32` now corrected as shown below:
```
bytes memory parameters = abi.encode(
    bytes32("1b"),
    bytes32("MyBytes32"), 0x1fd36c61981313c0c155d33ffac0325bd7c00d21d52442981bb13d2fa13e8f71,
);
```


Siegrift (approved):
🙏 Thanks. I missed this one when doing the docs for `string32` type.

PR #714, opened 2022-04-20
Remove validator from v0.6

Siegrift:
https://api3dao.atlassian.net/browse/AN-689


wkande (approved):
Looks good. I will place a copy of validator.md into archives with my next PR.

PR #721, opened 2022-04-24
Adjust language of why power beacons

dcroote:
I've also reorganized a bit to prioritize the sections I think API providers are most interested in.


wkande (commented):
Real nice thanks

PR #727, opened 2022-04-27
Test for CORS

wkande:
This PR will be discarded along with the branch behind it. Using it only for the firebase site.


wkande:
I build this PR so @aquarat could see my CORS issue for both localhost and on a firebase hosting site.

PR #730, opened 2022-04-30
Document validator

Siegrift:
https://api3dao.atlassian.net/browse/AN-608
and
https://api3dao.atlassian.net/browse/AN-609

The documentation are separated into two commits for better visibility.


Siegrift:
I've tested the documentation code snippets and examples by publishing locally to verdaccio.

wkande (commented):
This seems to be easier to use now. Let me know when it is ready to merge, thanks.

Siegrift:
@wkande It's ready to review and merge

wkande:
This is great as it now is simple to use and understand. I would like to return to it in the next few days with just a very minor change if that is OK.

PR #731, opened 2022-04-30
Remove duplicated config-json text

dcroote:
Looks like it originated from accidental duplication of `blockHistoryLimit` text


wkande (approved):
Good catch thanks.

PR #743, opened 2022-05-05
Added warning for gateways not to have the same API KEY

wkande:


Siegrift:
FYI: Validator will check this as well in the future.

PR #745, opened 2022-05-06
Add gasPriceMultiplier config setting for v0.7

vponline:
This adds an optional `gasPriceMultiplier` setting for Legacy gas prices (works similar to the `baseFeeMultiplier` but for legacy gas prices) for v0.7.


dcroote (approved):
LGTM 👍

PR #748, opened 2022-05-10
AN-727 Consolidate config.json chains info

dcroote:


docs/airnode/v0.7/grp-providers/guides/build-an-airnode/configuring-airnode.md
```diff
@@ -101,129 +103,6 @@ Below is a simple chain array with a single chain provider.
 ],
 ```
```
wkande, line 103: Why remove the entire chains section below? Will it be replaced with something else?

dcroote: There was a lot of duplication between the two files while at the same time neither was sufficient to be used alone, hence dev confusion on where to add content and user confusion with having to click back and forth. What I've done is consolidate all of the chains details into reference and provided a link to there from the configuring page.

wkande: I'm not sure about the approach to be honest. **Configuring an Airnode** explains parameters and how to use them. It is part of the **Build an Airnode** guide which steps through and details how to setup an Airnode. As part of the guide it should follow the guide's purpose and flow. The config.json is purely a reference section, quick defintions.

Your approach would thus sent the user to the reference section to understand each parameter. Are you thinking the reader will be more focused on the references > config.json than the Guide?

I understand the duplication issue. Maybe we should have a call to see what the Guide might do to better explain the  configuration of an Airnode with regards to the configuration page.

wkande:
Could we redefine the **Configure an Airnode** doc to simply and softly describe what each parameter does and then link to the **reference > config** page for a technical defintion?

docs/airnode/v0.7/grp-providers/guides/build-an-airnode/configuring-airnode.md
dcroote: But Configuring isn't detailed enough to enable a user to create the config.json (example `txType`) and hence they need to go to References anyway... My proposal is to have one source of truth rather than duplicated content.

wkande:
Ok so I am warming up to the idea. If **reference > config.json** did a better job of explaining each parameter (i.e. `gasPriceMultiplier` which confuses me to no end) we could create a list of links to each parameter in **reference >config**. What is bothering me is sending the reader somewhere else and then making them look for the parameter on the page. The links should take them directly to the parameter and the back button will them return to where they were in the **Configuring and Airnode** doc..

wkande:
I create a JIRA for this and assigned it to you.

dcroote:
What about this compromise: the Configuring page only describes fields that need to be configured. Anything with a stable default isn't discussed and is instead covered by a general link to that section (e.g. chains) within the Reference page.

wkande:
I would rather go with a list of links per section rather than the compromise.

So for `chains` in **Configuring an Airnode** there is a list of parameters that point to each parameter in the **reference > config.json.md** page. This could be before or after the code example for each root key of `config.json` in **Configuring an Airnode**. As an example I added an actual link to the `options` key below.

- [authorizers]()
- [contracts]()
- [id]()
- [providers]()
- [type]()
- [options](https://docs.api3.org/airnode/v0.7/reference/deployment-files/config-json.html#options)              <------- try this one
  - [options.txType]()
  - [options.priorityFee]()
  - [options.baseFeeMultiplier]()
  - [options.fulfillmentGasLimit]()
- [maxConcurrency]()
- [blockHistoryLimit]()
- [minConfirmations]()
- [ignoreBlockedRequestsAfterBlocks]()

dcroote:
I now think the idea behind this PR may not be so great. If a user wants to quickly try out Airnode they'll use a Quick Deploy tutorial. If they're reading the Configuring page then they want additional detail, in which case referring to a Reference page with a list of links isn't overly helpful. 

I still think there are a couple of opportunities to clarify some of the fields / language, but I think I'll close this PR and address those separately, unless you're opposed @wkande .

wkande:
I guess I don't understand your concern to rethink this but want to see what you have in mind. The opportunity to dedup the field (key) definitions is a big plus because I think it annoys developers who try to edit both these pages when they make a change. Of course they can be taught what each page tries to convey.

wkande:
> Preview site: https://api3-docs--dedup-config-json-upv2zj24.web.app

Looks good so far.

wkande:
So a new consideration arrises. I have been working on a new search component. Right now the search only filters by heading elements. If the config.json fields are not mentioned in a heading element and only as links, then search will not find them. So searching say for `blockHistoryLimit` would only find it in **reference > config-json-md**.

I was going to introduce tags for each page but the search would only take the user to the top of the page that has at the tag.

Adding the entire content of each page will get expensive (labor) by using something like Algolia. Maybe that becomes the next step to consider after the gen2 search I am building is completed.

dcroote:
@wkande  I think this is now ready to go- I've consolidated all of the config.json sections and confirmed the links work. Having done it I'm convinced this was the right call given there were minor bugs and pieces that had gotten out of sync.

wkande (approved):
OK let's give it. go.

PR #754, opened 2022-05-14
Fix comment for broken link

dcroote:
Follow-up to #751. Adds spaces to comment out the text and removes the link to avoid repeated link checker failure.


wkande (approved):
Guess I missed this one.

PR #759, opened 2022-05-18
Update request status concepts and configs

dcroote:
The process of assigning requests a status, including blocked and ignored, was removed from Airnode prior to v0.5 [via this commit](https://github.com/api3dao/airnode/commit/0ed6277bdd789bfa48d97e6c5d179c9ba357a520). As part of that `ignoreBlockedRequestsAfterBlocks` was also removed.


Siegrift (commented):
👍 Thanks.

PR #777, opened 2022-05-26
replace timestamp information for value only functions

Ashar2shahid:
small details I missed, these should be correct now

### DIFF
--- a/docs/beacon/v0.2/developers/read-data-feed-value-with-dapi-name.md
+++ b/docs/beacon/v0.2/developers/read-data-feed-value-with-dapi-name.md
@@ -73,8 +73,5 @@ this is an issue, consider casting the timestamps to a larger type.
 
 :::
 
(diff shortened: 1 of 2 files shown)


wkande (commented):
Change of thought here. Deleted my first response as it was way off. So the "value only" functions return an error message if the timestamp is zero, meaning the feed does not exist? Is the error message in the value?

wkande:
And, does a zero timestamp no longer have a zero value because an error is returned?

Ashar2shahid:
> Change of thought here. Deleted my first response as it was way off. So the "value only" functions return an error message if the timestamp is zero, meaning the feed does not exist? Is the error message in the value?

No, it reverts with that message


> And, does a zero timestamp no longer have a zero value because an error is returned?

afaik a zero timestamp means that the value will always be zero. However a zero value does not mean the timestamp is zero. For this function in particular we don't return the value and instead raise an error if the timestamp is 0

PR #779, opened 2022-05-28
Update Api3Voting.sol link

dcroote:
This contract was in a different directory than what the existing link points to.
~~Also fixes formatting of a previous commit.~~


wkande (approved):
looks good

PR #786, opened 2022-06-02
Clarify reserved parameters, fix minor inconsistencies

Siegrift:
https://api3dao.atlassian.net/browse/AN-616

### DIFF
--- a/docs/ois/v1.0.0/ois.md
+++ b/docs/ois/v1.0.0/ois.md
@@ -439,10 +439,11 @@ that cannot be overridden by the requester.
 ### 5.4. `reservedParameters`
 
 [<InfoBtnBlue/>](/airnode/v0.6/grp-providers/guides/build-an-airnode/api-integration.md#reservedparameters)
-(Optional) A list of objects that specify reserved Airnode endpoint parameters
-that do not map to any API operation parameters, but are used for special
-purposes by the Airnode. See the [Reserved Parameters](./reserved-parameters.md)
-doc for an in-depth explanation. Each object has the following elements:
+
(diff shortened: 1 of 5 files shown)


docs/ois/v1.0.0/ois.md
```diff
@@ -457,11 +458,13 @@ Allowed values: `_type`, `_path` or `_times`
 #### 5.4.2. `fixed`
 
 (Optional) The fixed (i.e., non-overridable) value for the reserved parameter.
+If `fixed` is used, the `default` has no effect and should not be used.
```
wkande: The extra use of the word "**the**" before "**default**" throws me off, maybe:

If `fixed` is used, `default` has no effect and should not be used.

docs/ois/v1.0.0/ois.md
```diff
@@ -457,11 +458,13 @@ Allowed values: `_type`, `_path` or `_times`
...
 
 #### 5.4.3. `default`
 
 (Optional) The default value for the reserved parameter. Used when no value is
-provided.
+provided. If `default` is used, the `fixed` must not be used because it would
```
wkande: Same here.

If `default` is used, `fixed` must not be used because it would...

dcroote (approved):
@Siegrift  - I added your fix to the previous `config.json` pages, hope that was OK

Siegrift:
>  I added your fix to the previous config.json pages, hope that was OK

Totally, thanks for that.

wkande (approved):
I did one tiny grammar change. Will merge now.

PR #787, opened 2022-06-02
Content reflective of data-feed-reader-example

wkande:
The only thing still outstanding is the canRead functions only accepting `_datafeedId`. There is a question out on Slack about it.

### DIFF
--- a/docs/data-feeds/developers/README.md
+++ b/docs/data-feeds/developers/README.md
@@ -17,12 +17,40 @@ The
 contract serves data feed values to any dApp with the appropriate permissions.
 The contract is simple to use and returns immediate data feed values.
 
-::: tip Data feed preview access
+## Starter Project
 
-Currently access to data feeds does not require a subscription during the
-current preview period.
+The
+[data-feed-reader-example](https://github.com/api3dao/data-feed-reader-example)
+starter is an example project for reading API3 data feeds on the Polygon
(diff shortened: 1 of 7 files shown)


bbenligiray:
We capitalize Beacon because it's the name of an API3 product, and isn't used semantically. Compared to that, when we say "data feed", we refer to the general, widely accepted meaning of that phrase. So "data feed" isn't a proper noun/shouldn't be capitalized.

There are some inconsistencies with capitalization related to this, such as
> [Developers - Why API3 Data Feeds?](https://api3-docs--bec186-20220602-n9y7669b.web.app/data-feeds/introduction/why-use-data-feeds.html)
> [API Providers - Why power Beacons?](https://api3-docs--bec186-20220602-n9y7669b.web.app/data-feeds/introduction/why-power.html)

("power" not being capitalized implies "data feed" was capitalized and this is not a header styling choice)

bbenligiray:
I updated https://github.com/api3dao/data-feed-reader-example to always prioritize dAPIs over Beacons/Beacon sets (though it was already like that), can you repeat that here? For example [here](https://api3-docs--bec186-20220602-n9y7669b.web.app/data-feeds/), the reader needs to read too much until they reach dAPIs. Similarly, the Reference section on the sidebar has a lot about Beacons and not enough about dAPIs.

PR #789, opened 2022-06-05
AN-761 Chain idiosyncrasies

dcroote:
[AN-761](https://api3dao.atlassian.net/browse/AN-761)


wkande (approved):
Looks good to go, thanks.

PR #801, opened 2022-06-08
Document processing updates

Siegrift:
https://api3dao.atlassian.net/browse/AN-750


wkande (approved):
I will merge this and make some minor layout changes if that is OK. By using heading 2 captions you two tips will get better exposure in the search.

PR #811, opened 2022-06-13
Add config template docs

vponline:
This adds documentation for the new `templates` field in `config.json` which allows hardcoded templates (V0/RRP only) to be used by the Airnode.


wkande (approved):
Looks good. I may make a small change to the configure an Airnode page later. Each sub-field should have a bullied description that in turn links to the reference section. This makes things consistent with the other listings on the page.

PR #858, opened 2022-06-30
Change "oisFormat" to "1.1.1"

### LINKED ISSUE #856: Use OIS package from separate repository in Airnode
Documentation for https://github.com/api3dao/airnode/issues/1274

The `oisFormat` version changed to `1.1.1` and needs to be updated in all occurences.

Siegrift:
Closes https://github.com/api3dao/api3-docs/issues/856

### DIFF
--- a/docs/airnode/v0.8/grp-providers/guides/build-an-airnode/api-integration.md
+++ b/docs/airnode/v0.8/grp-providers/guides/build-an-airnode/api-integration.md
@@ -85,7 +85,7 @@ json object.
 
 ```json
 {
-  "oisFormat": "1.0.0",
+  "oisFormat": "1.1.1",
   "title": "myOisTitle",
   "version": "0.1.0",
   ...
@@ -94,7 +94,7 @@ json object.
 
 ### oisFormat
 
-Leave this as `1.0.0`, which is the current OIS format version.
+Leave this as `1.1.1`, which is the current OIS format version.
 
 ### title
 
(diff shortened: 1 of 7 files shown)


wkande:
@Siegrift : If Airnode version `v0.8` is going to use OIS `v1.1.1` I assume that I need to create a `v1.1.1` version of the OIS doc set?

Siegrift:
That is a good point. I think we should just have major/minor version similarly how we do for Airnode.
So this means we should change `1.0.0` label to `1.0` and introduce a new version `1.1`.

Also, https://github.com/api3dao/api3-docs/commit/d4fe68cce96bb2d74e9ff8a7c20c67d8e763fdd8 should be part of `1.1` and removed from `1.0`.

Siegrift:
@wkande Do you need my help to make the changes?

wkande:
@Siegrift Looks like Airnode `v0.8` now uses OIS `v1.1.1` correct? So I need to create a new OIS version `v1.1.1`. This implies we are not publishing a OIS `v1.1.0`, or are skipping it. If this is correct then we will need to go into the OIS doc set for `v1.1.1` and update it as well with changes.

Siegrift:
We should ignore the patch. Only consider major and minor versions. We should have:
- `v1.0` - used for Airnode versions <= 0.7
- `v1.1` - used by Airnode 0.8

wkande:
I would agree. Burak thought to use `v1.0.0` which is what we have now. Let's bring this up in the call tomorrow. It would be simple to change to `v1.0` and `v1.1`.

wkande (approved):
Merging this today. Will create a new issue to add a new version to the OIS doc set.

PR #861, opened 2022-07-01
Fix link checker CI by disabling a failing URL link

Siegrift:


dcroote (changes requested):
The CI still failed here because the link should be added to the [lychee.toml file](https://github.com/api3dao/api3-docs/blob/7240e3e7174138c6adb9a089ea0b759460d9bd70/.github/workflows/lychee.toml#L56) exclude array rather than through this comment strategy.

wkande:
I will update the [lychee.toml file](https://github.com/api3dao/api3-docs/blob/7240e3e7174138c6adb9a089ea0b759460d9bd70/.github/workflows/lychee.toml#L56).

Issue #869, opened 2022-07-05
Add OIS `v1.1` doc set


wkande:
We seem to need a new version for the OIS doc set. Should it be `v1.1.1` or `v1.1` doc set. See the discussion [here](https://github.com/api3dao/api3-docs/pull/858).

OIS versions:
v1.0 - used for Airnode versions <= 0.7
v1.1 - used by Airnode 0.8

wkande:
This will involve refactoring `v1.0.0` to `v1.0`.

wkande:
PR: https://github.com/api3dao/api3-docs/pull/876

martinkolenic:
Hi @wkande, I looked at this earlier in the day and it looks just fine to me, both v1.0 and v1.1 are accessible and I have no complaints content-wise. Moving this to Done then.

PR #876, opened 2022-07-06
869 - new OIS doc set version `v1.1`

wkande:
Closing #869

Refactor `v1.0.0` to `v1.0` and created `v1.1`.

@Siegrift and @amarthadan - Please check out the preview for this PR. Is this the way we want to version the OIS doc set. I can change it back very easily. Also the new doc set for OIS `v1.1` is a copy of `v1.0` at this point. I assume there are changes but I cannot find a CHANGELOG. Can someone go into the OIS doc set for `v1.1` and update anything relevant?


wkande:
I'm wondering if the version pick list in the top navigation bar would be better illustrated using the following.
- v1.0.x
- v1.1.x

In fact this may be the case for all doc sets that are versioned.

dcroote (commented):
There is actually only a [single difference](https://github.com/api3dao/ois/compare/v1.0.0...v1.1.1#diff-cc7ae7de5a5b943e6f4108ec9f055e986a13564aef3d342d1d5d902c237c900bR123): the addition of `relayRequestId`. Consequently if you could add that to [this list](https://api3-docs--869-new-ois-docset-version-c1gg62j7.web.app/ois/v1.1/ois.html#_4-3-1-type) I think that would do it.

Also I don't see `v1.1` in the preview site top navigation bar pick list, though I can manually go to v1.1 by adjusting the URL, not sure if that is expected.

>I'm wondering if the version pick list in the top navigation bar would be better illustrated using the following.
v1.0.x
v1.1.x
In fact this may be the case for all doc sets that are versioned.

I see the point, but currently Airnode docs use the major / minor only e.g. `v0.7` despite having the patch version used within configs and whatnot, so I would say `v1.0` and `v1.1` of the OIS docs are fine, but don't feel particularly strongly.

Siegrift:
> There is actually only a [single difference](https://github.com/api3dao/ois/compare/v1.0.0...v1.1.1#diff-cc7ae7de5a5b943e6f4108ec9f055e986a13564aef3d342d1d5d902c237c900bR123): the addition of relayRequestId. Consequently if you could add that to [this list](https://api3-docs--869-new-ois-docset-version-c1gg62j7.web.app/ois/v1.1/ois.html#_4-3-1-type) I think that would do it.

We should mention this only in the v1.1 docs sets and omit that from the v1.0

wkande:
@dcroote @Siegrift If all that was added was a `component.type` of `relayRequestId` then why did we jump to `1.1.1`?

wkande:
> Also I don't see `v1.1` in the preview site top navigation bar pick list, though I can manually go to v1.1 by adjusting the URL, not sure if that is expected.

Yes, OIS `v1.1` will be published when Airnode `v0.8` is published. Until then both are missing from their version pick lists.

Siegrift:
> If all that was added was a component.type of relayRequestId then why did we jump to 1.1.1?

Up until now, we didn't really version things. We always changed the implementation without changing the version number. This was mainly for convenience reasons. Recently, we moved OIS to a separate repo and the versioning is based on a `package.json` version. Since, npm version releases are immutable, we are forced to bump a `package.json` version before doing a release. 

Addition of `relayRequestId` is a minor change (according to the semver rules) so thats why we published version `1.1.0`. However, the release was broken (due to technical error) so we fixed it and released `1.1.1`. We will honor semantic versioning for OIS from this point on.

Finally, Airnode is always bound to a single version of OIS, which is based on the version of the dependency installed. See https://github.com/api3dao/airnode/blob/master/packages/airnode-node/package.json#L30. Airnodes with versions <= 0.7.x use OIS version 1.0.0 and the new (not yet released) airnode@0.8 version will use `1.1.1`. (We might to note this somewhere in the docs).

wkande:
Sounds good. I did add mention of the OIS > Airnode compatibility **[here](https://docs.api3.org/ois/v1.0/#releases)**.

PR #880, opened 2022-07-07
Fix minor typos related to the autoremove function

martinkolenic:
This PR only corrects a couple minor typos introduced with the description of the autoremove feature.


wkande (approved):
Good catch, thanks

PR #886, opened 2022-07-08
Add doc set for /operations

### LINKED ISSUE #885: Add operations doc set
The operations doc set will be like the dev doc set in that it does not have a presence in the doc set pick list on top of the sidebar. Also update the document set Vue component to hide the pick list when either `/dev` or `/operations` is the current doc set.

wkande:
Closes #885 
@aquarat @kolenic-martin 

There is now an `/operations` path in the docs. This opens up a new **Operations docs set**. The doc set will not be loaded into the doc set pick list which is normally presented at the top of the sidebar. In addition, the pick list is hidden when the url contains `/operations`. This is the same behavior as `/dev`. 

The doc set is public and can be discovered but this could be made private if needed I suppose.

Please check out the **[preview site here](https://api3-docs--885-operations-doc-set-04pozze3.web.app/operations/)**.

### DIFF
--- a/docs/.vuepress/components/DocumentSets.vue
+++ b/docs/.vuepress/components/DocumentSets.vue
@@ -1,13 +1,17 @@
 <!--
 This component places document sets in the header of the sidebar. 
 
-NOTE: When this component is MOUNTED (as it is mounted from a landing page click),
+NOTE: When this component is MOUNTED (as it is mounted from a landing page click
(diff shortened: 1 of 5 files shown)


martinkolenic:
Hi @wkande, the preview works for me, perhaps consider making it private while it is empty just for good measure?

PR #891, opened 2022-07-12
Update aws.env file handling instructions and docker commands

vponline:
This updates the `aws.env` credentials to be mounted into the docker image as a file (and then exported from there by the deployer) instead of being read directly by docker into the container. env variables This allows the `aws.env` file to contain double quotes.

PR #888, opened 2022-07-11
Split deployer remove command in two and output receipt.json to config/

dcroote:
Documentation for api3dao/airnode#1322.

Rather than one clunky `remove` command, there are now two remove commands.  `remove-with-receipt` (preferred) uses `receipt.json` while `remove-with-deployment-details` expects the various deployment details instead.

The `receipt.json` default output folder is now `config/` rather than `output/` and therefore all references to the `output/` configuration directory have been removed or replaced, including in the `deploy` command. 

Closes #855

PR #894, opened 2022-07-15
Remove duplication in deployment instructions

dcroote:
This PR will be rebased after #891 is merged.

After #888 I noticed there was significant duplication in a couple of deployer-related pages. This PR removes that duplication without detracting from the user flow. For example, in the guide for deploying an Airnode, the example commands are still present, while the deployer docker image page now has a reference to that section rather than all of those commands duplicated. Similarly, the docker image page is now the sole source of information on removing an airnode, rather than have those commands also be present in the deployment guide.

In doing so, this also addresses [this comment](https://github.com/api3dao/api3-docs/pull/888#issuecomment-1184118294) by @kolenic-martin .

### DIFF
--- a/docs/airnode/v0.8/grp-providers/docker/deployer-image.md
+++ b/docs/airnode/v0.8/grp-providers/docker/deployer-image.md
@@ -97,8 +97,8 @@ deployment information and is used to remove the Airnode.
 
 ```sh
 docker run -it --rm \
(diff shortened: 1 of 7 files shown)


martinkolenic:
Thanks, @dcroote, this LGTM.

wkande (approved):
Again, looks good.

PR #895, opened 2022-07-15
Remove legacy/eip1559 tx options, add providerRecommendedEip1559 strategy

vponline:
This removes EIP1559/Legacy chain transaction options and adds `providerRecommendedEip1559GasPrice` as a new strategy, I also fixed some of the json formatting that was off.

### DIFF
--- a/docs/airnode/v0.8/concepts/chain-providers.md
+++ b/docs/airnode/v0.8/concepts/chain-providers.md
@@ -51,33 +51,35 @@ Airnode. It then applies an arbitrary name for the blockchain provider
     },
     "type": "evm",
     "options": {
-      "txType": "eip1559",
-      "priorityFee": {
-        "value": 3.12,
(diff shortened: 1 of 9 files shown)


wkande (approved):
I will merge this. Later (tomorrow I hope) I plan to make an adjusted. When completed perhaps you might review the change for me. See https://github.com/api3dao/api3-docs/issues/896

In the file `reference/deployment-files/config-json.md` we need to relocated the `strategies`: latestBlockPercentileGasPrice, providerRecommendedGasPrice and constantGasPrice to be under `gasPriceOracle` as possible strategy values.

dcroote:
@vponline would you be able to update the [Chain Idiosyncrasies](https://docs.api3.org/airnode/v0.8/reference/chain-idiosyncrasies.html) page with these new gas strategies?

PR #897, opened 2022-07-16
Replace encode() with abi.encodePacked()

wkande:
In code example`/concepts/tempplates.md`.


wkande:
@bbenligiray The correction for `encodePacked()`.

PR #898, opened 2022-07-16
Restructure the QRNG docs

### LINKED ISSUE #881: QRNG docs assume an EOA sponsor, while qrng-example has the requester sponsor itself
See https://discord.com/channels/758003776174030948/765618225144266793/994593770509189131

In the context of qrng-example, sponsor and requester addresses are the same for simplicity. The docs have the user generate a sponsor mnemonic, which is not needed. This may be why QRNG users have been confused about the sponsorship scheme. The docs need to be updated to match qrng-example. Alternatively, strip the docs down and rely on qrng-example more heavily.

wkande:
This commit represent a complete restructure of the QRNG docs. A new section called Guides has been added. Here it was important to add the QRNG Example Project as the first guide as it needed to be easier to find. Two other guides have been added.

Closes #881


Ashar2shahid:
Looks good, good job 👍

Ashar2shahid (approved):

PR #924, opened 2022-07-25
Adding workspaces to the side bar and fixed a few grammar mistakes

RyanMakiola:
1. Added workspaces to the side bar
2. fixed a few grammar mistakes


wkande (approved):
Looks good @RyanMakiola  will reply this right away.

PR #936, opened 2022-07-27
Fix signed data gateway response format

### LINKED ISSUE #934: Fix incorrect signed data gateway response format
See: https://api3dao.atlassian.net/browse/BEC-357?focusedCommentId=17080

Just to clarify. This is only a docs change. The implementation works as intended.

Siegrift:
Closes https://github.com/api3dao/api3-docs/issues/934


wkande (approved):
Seems simple enough. Thanks for the update.

Issue #873, opened 2022-07-05
Search effectiveness on multiple headings


wkande:
The `match-query.js` script used by VuePress does not pick up the second occurrence of a filter term when it appears in both the H1 and lower headings (H2, H3, e.t.c). If the term is in multiple headings other than H1, it will find both.

wkande:
PR: https://github.com/api3dao/api3-docs/pull/939

PR #939, opened 2022-07-30
873 search multiple headings

wkande:
The `match-query.js` script used by VuePress does not pick up the second occurrence of a filter term when it appears in both the H1 and lower headings (H2, H3, e.t.c). If the term is in multiple headings other than H1, it will find both.

`match-query.js` has been replaced will an easy to understand function that will allow for new features in the future.

The results layout has a minor change.


wkande:
Closes #873

PR #943, opened 2022-08-01
ChainAPI: Add a section to workspaces about the activity log

rlt:


wkande (approved):
Thanks for the update, will merge now.

PR #944, opened 2022-08-01
Format Windows code blocks as batch not bash / sh

dcroote:
Replaces `sh` and `bash` with `batch` for syntax highlighting of Windows CMD code blocks and with `powershell` for the few Powershell code blocks. Prismjs supported languages [are here](https://prismjs.com/#supported-languages) as a reference.

See the preview site below for how it renders.


wkande (approved):
Nice to have a "Windows" dude to catch these thing. Thanks. Will merge in about 2 hours or so with other PRs.

PR #956, opened 2022-08-05
Clarify cloudProvider.region in config.json

### LINKED ISSUE #950: Test GCP region and zones
Run the GCP tutorial using just a region `us-east1` and using a region/zone `us-east1-a`. Then determine the correct terminology for the docs [**here**](https://docs.api3.org/airnode/v0.7/reference/deployment-files/config-json.html#cloudprovider-region).

dcroote:
Closes #950.

`region` and not `zone` as confirmed by @vponline.

Preview site: https://api3-docs--config-provider-region-fix-2kymxgi1.web.app/


wkande (approved):
@dcroote - I see that you removed the link to Google that explains the difference between a region and a zone. I will merge this but please let me know if you feel the link should be added back in.

dcroote:
Thanks, I think we're safe skipping the link.

PR #975, opened 2022-08-16
Fix invalid reference

bdrhn9:
Just modified only for version 0.7 and 0.8. If it needs, I can also change for previous ones.


wkande (approved):
Good catch thanks

PR #1007, opened 2022-09-05
updated deployments docs page

RyanMakiola:
updated the deployments docs page to incorporate all the new changes we have made to the deployment flow during the network settings v2 tasks


wkande (approved):
Merging

wkande:
I will fix the Prettier formatting when I reply this PR with some other as part of a daily correction.

PR #1021, opened 2022-09-10
Add deployment data to heartbeat

### LINKED ISSUE #1004: Add deploy data to heartbeat payload
Documentation for [1411](https://github.com/api3dao/airnode/issues/1411)

vponline:
Closes #1004


wkande (approved):
@vponline Again the odd thing here is that the heartbeat ID was already gone from my local repo from a previous merge of your early PR. So double checking the deployment to docs.api3.org might be a good thing. I will check it right after the deployment and if you could as well so we know everything is OK I would appreciate it, thanks.

vponline:
> @vponline Again the odd thing here is that the heartbeat ID was already gone from my local repo from a previous merge of your early PR. So double checking the deployment to docs.api3.org might be a good thing. I will check it right after the deployment and if you could as well so we know everything is OK I would appreciate it, thanks.

Also checked the v0.10 docs page here and the data is there correctly :+1:

PR #1022, opened 2022-09-10
Make simultaneous deployments warning v0.8+ compatible

dcroote:
This change makes the warning compatible with `v0.8` where the `remove` command has been replaced with `remove-with-receipt` and `remove-with-deployment-details`.


wkande (approved):
Sounds good thanks

PR #1040, opened 2022-09-19
Remove outdated tx considerations section

dcroote:
The addition of gas price oracle strategies rendered this section obsolete. As one example, `txType` no longer exists.


wkande (approved):
I guess this one got missed. Thanks.

PR #1041, opened 2022-09-19
ChainAPI: Add DNS verification section to Workspaces

rlt:


RyanMakiola:
@rlt looks good, we could look at adding that it refreshes every hour, however because we mention this within our UI I dont feel strongly about mentioning the hourly refresh in our docs.

rlt:
Yeah as per the discussion on that JIRA, I'm not sure we should be mentioning how often we check at all, as that ties our hands if we need to adjust the schedule over time.

PR #1043, opened 2022-09-19
chains[n].authorizers should be an empty object, not an empty array

andreogle:
I was going through the docs for v0.8 and noticed this issue with `chains[n].authorizers` referring to an empty array, rather than an empty object (the new structure). I hope this covers all of the references


wkande (approved):
I appears to be messed up in two other places as well. I will merge your changes and follow up getting the following pages fixed.

- https://docs.api3.org/airnode/v0.8/concepts/authorization.html
- https://docs.api3.org/airnode/v0.8/grp-providers/guides/build-an-airnode/apply-auth.html

PR #1068, opened 2022-10-03
Add removal via deployment ID

### LINKED ISSUE #1064: Allow users to remove the deployment based on the deployment ID
Documentation for https://github.com/api3dao/airnode/issues/1460

amarthadan:
Close #1064

I updated the docs so the new `remove` command is the preferred way to remove Airnodes. Just want to double-check with @bbenligiray that this was the intention.


bbenligiray:
Sounds good to me for this to be the recommended way. Is it worth it to mention that they should use the same deployer version?

wkande (approved):
Looks good to me. Thanks

PR #1070, opened 2022-10-04
added import & export to chainapi documentation

RyanMakiola:


wkande:
@RyanMakiola AND @rlt : I am merging this now. It does have "**Prettier**" formatting issues which I will fix for you before publishing the changes. You can check if **Prettier** formatting needs to be executed by running the following script. Do so before pushing your changes to the repo. 

```
// Change to the root of api3-docs
cd <root-of-api3docs-folder>

// Check if formatting is needed.
yarn run format:check

// If needed run:
yarn run format
```

```j
// The two scripts come from the packages.json file
"scripts": {
    ...
    "format": "prettier --write \"./**/*.{js,vue,md,json,yaml}\" --loglevel silent",
    "format:check": "prettier --check \"./**/*.{js,vue,md,json,yaml}\"",
    ...
  },
```

dcroote:
> You can check if Prettier formatting needs to be executed by running the following script. Do so before pushing your changes to the repo.

If you run `yarn install` followed by `yarn prepare` in the top-level of the repo locally it will install a `pre-push` hook that will check formatting each time before you push 😁 

Also, I quite like is the Prettier extension for VS Code that automatically formats on save.

rlt:
Hey both. Thanks for the notes. I had a lot of trouble previously getting prettier to work on Ryan's set up (Aaron also helped out and struggled). I'll try give it another go. 

@wkande in future if Ryan assigns me as a reviewer, please let me approve before merging as I need to check the content of what is written.

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -0,0 +1,30 @@
...
+# {{$frontmatter.title}}
+
+<TocHeader />
+<TOC class="table-of-contents" :include-level="[2,3]" />
+
+Importing and exporting API integrations allows for easy sharing of integrations across workspaces, making integration backups and other use cases.
+
+## Import OIS
+
+When starting a new integration, you have the option to click “Import OIS” which allows you to paste an integration OIS ([click here](https://docs.api3.org/ois/v1.1/) for more information) that will be used to pre-populate the integration wizard for you. You will then be able to complete the integration wizard.
```
rlt, line 20: - Say `adding` instead of `starting`

- `integration OIS` isn't right because the I stands for integration - so just say OIS.

- Let's not call it `wizard`, that sounds quite dated and we haven't used that elsewhere. No need to replace it, just drop the term altogether.

- Is there a way to link to latest rather than a specific version of OIS?

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -0,0 +1,30 @@
...
+---
+
+<TitleSpan>{{$frontmatter.folder}}</TitleSpan>
+
+# {{$frontmatter.title}}
+
+<TocHeader />
+<TOC class="table-of-contents" :include-level="[2,3]" />
+
+Importing and exporting API integrations allows for easy sharing of integrations across workspaces, making integration backups and other use cases.
```
rlt, line 16: This isn't right, we're not importing and exporting _integrations_, only the OIS portion of them. This is important because a lot of details don't make it across, such as the endpoint documentation.

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Import OIS
+
+When starting a new integration, you have the option to click “Import OIS” which allows you to paste an integration OIS ([click here](https://docs.api3.org/ois/v1.1/) for more information) that will be used to pre-populate the integration wizard for you. You will then be able to complete the integration wizard.
+
+* If you are importing the OIS for a new integration using a config.json, you will only be able to import an object from the OIS array section from the config.json
```
rlt, line 22: This is tricky to read - rather tell someone _how_ to do what they need to and not a restriction (|you will only be able") - it comes back to negative tenses we've chatted about before. Change to something like `If you are using a config.json file, copy the OIS object from within the array and paste into in the text area provided` and provide a code example of the section to copy too.

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -0,0 +1,30 @@
...
+* Importing an OIS is only available for new integrations and you are not able to import an OIS for an existing integration
```
rlt, line 23: - This point should come first and its redundant to say you can't do it for an existing integration as you've said its only available for new integrations.

- Both bullets should be wrapped in with a `Important Notes` header or similar

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+## Export OIS
+
+You can export an integration OIS from one of the below sections:
```
rlt, line 28: drop integration from `integration OIS`

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -0,0 +1,30 @@
...
+* Navigate to your integrations, click "Manage Versions" on the integration you want to export. Then click the ellipsis on the right of the version you wish to export and select "Export OIS"
+* From the "Manage Versions" page above, you can select "View" for the version you want to export. You will then see an "Export OIS" button on the top right of the page
```
rlt, line 30: When providing steps, give each one a heading and number the points. Long bullets are really hard to read.

```
Manage Versions
1. Navigate to your integrations
2. Click on Manage Versions
etc
```

rlt (commented):
There's a few things to address here. On a higher level, we've been inconsistent in the docs on whether actions should be bold or in quotes. The integration and deployments pages use bold, workspaces quotes. I would use bold across the board.

wkande:
> @wkande in future if Ryan assigns me as a reviewer, please let me approve before merging as I need to check the content of what is written.

@rlt - @RyanMakiola  The best way to do this is to require a certain reviewer, that way I cannot merge until reviewed.

docs/chainapi/user-guides/Import&Export.md
RyanMakiola: True, what do you think of the below?
```
Importing and exporting the OIS of your API integrations allows for easy sharing of integrations
across workspaces, making integration backups and other use cases.
```
or
```
Importing and exporting allows for easy sharing of integrations
across workspaces, making integration backups and other use cases.
```

rlt: Yeah I think the latter is better.

docs/chainapi/user-guides/Import&Export.md
RyanMakiola: For the last part, is it fine if we link them to https://docs.api3.org/ois/v1.1/example.html? 
because the actual code is a lot.

I was thinking something like this:
```
If you are using a config.json file, copy the OIS object from within the array and paste 
into the text area provided (click here [above link] for an example OIS object).
```

rlt: Yeah the link is fine - but use this instead of a specific version: https://docs.api3.org/ois/latest/example.html

Sentence looks good 👍🏻

PR #1075, opened 2022-10-06
Remove ropsten, rinkeby, kovan from chains

bbenligiray:
QRNG no longer responds to ropsten, rinkeby, kovan requests so we should remove these. We can't add sepolia here until we have AQN redeploy their Airnode and that is blocked by ChainAPI supporting publicly accessible deployments (it deploys with authorizers by default).


wkande (approved):
Added issue #1076 as a reminder and task to add sepolia.

PR #1078, opened 2022-10-06
Chainapi-import&export-updates

RyanMakiola:
Made additional changes locally based on https://github.com/api3dao/api3-docs/pull/1070

### DIFF
--- a/docs/chainapi/user-guides/Import&Export.md
+++ b/docs/chainapi/user-guides/Import&Export.md
@@ -18,10 +18,10 @@ workspaces, making integration backups and other use cases.
 
 ## Import OIS
 
-When adding a new integration, you have the option to click “Import OIS” which
-allows you to paste an OIS ([click here](hhttps://docs.api3.org/ois/v1.2/) for
-more information) that will be used to pre-populate the integration for you. You
-will then be able to complete the integration.
+When adding a new integration, you have the option to click **Import OIS** which
+allows you to paste an OIS for more information) that will be used to
+pre-populate the integration for you. You will then be able to complete the
+integration. Read more about OIS [here](https://docs.api3.org/ois/v1.2/).
 
 ::: tip Important Notes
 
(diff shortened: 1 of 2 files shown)


RyanMakiola:
latest preview link: https://api3-docs--chainapi-import-export-updates-cjldhdjp.web.app/

wkande:
@rlt will wait for you to approve this PR before merging. I am doing a production release in a few hours.

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -13,30 +13,40 @@ tags:
...
 
 ## Import OIS
 
-When starting a new integration, you have the option to click “Import OIS” which
-allows you to paste an integration OIS
-([click here](https://docs.api3.org/ois/v1.1/) for more information) that will
-be used to pre-populate the integration wizard for you. You will then be able to
-complete the integration wizard.
+When adding a new integration, you have the option to click “Import OIS” which
+allows you to paste an OIS ([click here](hhttps://docs.api3.org/ois/v1.2/) for
```
rlt: There's a typo here in the url. I wonder if we shouldn't have this in the opening section which is something like "Read more about OIS **here**" instead of hidden in this paragraph.

rlt (changes requested):
Based on Heikki's comments from earlier, we should probably mention something in the Export section around the difference between copying and downloading.

The actions should be in bold and not in quotes, like on integrations and deployments (and separately we should fix Workspaces)

docs/chainapi/user-guides/Import&Export.md
RyanMakiola: I think its useful having the link straight after where we mention it, but I think it would be nice to have it in an opening section too for people who later on want to recheck the OIS docs but dont want to go through the docs to find the link. What about having the same link in the opening section and in the paragraph where it is referenced? I have come across documentation that does both and thought it was a nice touch, but if I had to pick one implementation I think the opening section would be best. Another option could be adding a "important links" note that is in padding like I did for important notes, and we have all the links there?

rlt: I don't agree - by that logic we'd need to include this link in every section of this document where we mention OIS for the first time. Why does it merit a link here but not in the export section? This would get messy quickly. It's also out of context - I tripped over it when I reviewed. It implies you're going to learn more about pasting an OIS, but instead the link takes you to learning about OIS.

I also don't agree about adding an important links bit to every section. This should be once per page, that's pretty common across docs I've read. You could have a read more section in the opening paragraph or at the bottom of the doc.

RyanMakiola: Happy we go that route then. 
I have made the updates @rlt.

RyanMakiola:
@rlt just made another change after looking at the review link. Looks fine now for a review.

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -13,30 +13,48 @@ tags:
...
+4. Select **Export OIS**.
+
+**View Version page**
+
+1. Navigate to your integrations by clicking **Integrate** on the sidebar.
+2. Click **Manage Versions** on the integration you want to export.
+3. Select **View** for the version you wish to export.
+4. Click **Export OIS** on the top right of the page.
+
+When exporting, what is the difference between **Download OIS** and **Copy**?
```
rlt: This question style is a bit out of place with the rest of the docs. You could say `In the modal that appears, you have two options to export your OIS:`

docs/chainapi/user-guides/Import&Export.md
```diff
@@ -13,30 +13,48 @@ tags:
...
 
 ## Import OIS
 
-When starting a new integration, you have the option to click “Import OIS” which
-allows you to paste an integration OIS
-([click here](https://docs.api3.org/ois/v1.1/) for more information) that will
-be used to pre-populate the integration wizard for you. You will then be able to
-complete the integration wizard.
+When adding a new integration, you have the option to click **Import OIS** which
+allows you to paste an OIS for more information) that will be used to
```
rlt: `for more information)` should be removed here

rlt (commented):
Two minor updates but otherwise this is good to go 👍🏻

rlt (approved):

RyanMakiola:
@wkande this has been approved and is good to be merged.

PR #1090, opened 2022-10-17
Various fixes

aquarat:
This PR adds the safe navigation to the code that loads beacons, thereby preventing an error for empty providers.
This PR also adds the `engines` config field to `package.json`, which prevents someone with NodeJS 16 from trying to build this (it will fail to build).
During development the `docs:dev` command downloads resources which `prettier` tries to check for prettiness. This is incorrect behaviour and thus I have added this resource's path to the `.prettierignore`file.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -27,3 +27,5 @@
 /docs/airnode/v0.3/assets
 /docs/airnode/v0.3/figures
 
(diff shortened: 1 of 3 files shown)


wkande (approved):
Thanks for fixing this.

PR #1093, opened 2022-10-18
Reworked reserved parameters and added multiple reserved params

rlt:
- Reworks reserved parameters
- Adds light detail around multiple reserved parameters
- Minor updates to other sections and updates to heading indentations

### DIFF
--- a/docs/chainapi/user-guides/integrations.md
+++ b/docs/chainapi/user-guides/integrations.md
@@ -59,20 +59,21 @@ use when calling the any API operations.
 ### Base URL
 
 This is the first part of the URL to which all requests are made. The full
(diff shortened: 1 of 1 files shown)


RyanMakiola:
Your changes look good.

Outside of your changes, we currently have a "(?)" in the paragraph which is most likely a typo or a comment from the original documentation to check that sentence and we should remove it, maybe we can improve the wording too.

`The paths are captured by various (?) endpoints which we’ll get to in a second.`
to something like:
`The path is captured as part of each endpoint you add, we will expand on this under **New Endpoint**.`

RyanMakiola (approved):
Looks good to go.

PR #1096, opened 2022-10-19
Fix authorizations code example typo

dcroote:
`requesterEndpointAuthorizations` instead of `requesterEndpointAuthorizers` for the authorizations code example.


wkande (approved):
Another good catch, thanks

PR #1112, opened 2022-11-10
Add Pre & Post Processing to ChainAPI

rlt:


RyanMakiola (commented):
Just to double check, was naming it "Advanced Settings" intentional?

Reason I ask is the other headings are in line with what you see on ChainAPI, however on ChainAPI its just called "Advanced" where in the docs we refer to it as "Advanced Settings".

That said, I actually like it as advanced settings because it feels more explicit, makes me wonder if we should have it as Advanced Settings on the frontend.

Also, something unrelated to what your PR The first link on the integrations page is expired. Which leads to your concerns you initially had around having it a manual update. Do we want to update them when we make changes like this or are we going to do a periodic review of the links?

rlt:
While the docs site allows you to use `/latest/` in the URL to redirect to the latest version, when I've done this before, the lychee checker here breaks, I assume because the response is a redirect rather than a 200. Perhaps @wkande knows of a way around this? If we can use latest then we don't need to link to a specific version or check periodically.

I used Advanced Settings because Advanced feels a bit unfinished, it makes me thing "Advanced what?". Could be Advanced Features too.

RyanMakiola (approved):
This is good to go, I think using **Advanced Settings** for the title is the nicer option.

PR #1113, opened 2022-11-14
Fix QRNG Maximum Value & Minor Typo

bdrhn9:
I'm working on QRNG migration recently and saw mistake with maximum value of size parameter that QRNG uint256[] endpoint takes. This PR fixes the issue.


wkande (commented):
Thanks for the corrections

PR #1120, opened 2022-11-22
Clarify airnode-client provider URL when running locally

### LINKED ISSUE #1119: Clarify what the provider URL should be when running airnode-client against a local chain
See the screenshot from #dev-support below. I suppose the issue is current docs recommend `host.docker.internal` while `http://host.docker.internal:8545` is needed.
Screenshot from 2022-11-21 11-55-02

dcroote:
Closes #1119.

### DIFF
--- a/docs/airnode/v0.10/grp-providers/docker/client-image.md
+++ b/docs/airnode/v0.10/grp-providers/docker/client-image.md
@@ -128,9 +128,9 @@ docker run --detach ^
 
 > If you want to connect Airnode to a blockchain running on localhost, you need
 > to make the blockchain accessible from within the docker itself. If you use
-> docker for linux you can use `--network="host"` parameter. For windows, wsl or
-> mac connect to `http://host.docker.internal:8545` instead of
-> `http://127.0.0.1:8545`. See
+> docker for linux you can use `--network="host"` parameter. For windows, wsl,
+> mac, and potentially some flavors of Linux like Pop!\_OS, connect to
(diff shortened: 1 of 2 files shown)


peptide-origami:
Hey @dcroote, hope it's ok if I jump in here with a suggestion.

My OS is PopOs linux and I wasn't able to get the host-docker networks connected unless I used  `http://host.docker.internal:8545` in my airnode config in addition to `--network='host'` on the docker launch. As of [62a52aa](https://github.com/api3dao/api3-docs/pull/1120/commits/62a52aa246f19ece7160086de64d3ffcae808492), the document text reads: For windows, wsl or mac connect to `http://host.docker.internal:8545` instead of `127.0.0.1`. But I think it might need to be changed to specify this instruction should be used for all platforms. 

What do you think? Or are other linux users able to specify 127.0.0.1 in their config & get a successful launch?

dcroote:
Hi @peptide-origami, thanks for the feedback. We actually use `http://127.0.0.1:8545/` in our  GitHub Actions CI [end to end testing](https://github.com/api3dao/airnode/blob/6ade99798d832543cb64a03ef737be4c2accd94f/packages/airnode-examples/test/e2e/coingecko-local.feature.ts) which runs Linux. Still, I'll change the wording to mention that even some Linux variants might need to use `http://host.docker.internal:8545` as you describe.

dcroote:
@wkande no rush, but this is ready to merge now

wkande (approved):
Thanks

PR #1122, opened 2022-11-24
Rename sign-up to registration and include login options

rlt:


RyanMakiola (commented):
Looks good, just two minor changes to sections that were already there:

Under the create an account section, I think we should remove the "In the future" part.

Under "4. Register Email" I think we should remove "Enter your email, wait a moment for it verification, then select the Confirm Email button." it doesn't read well and feels like we over explaining

wkande:
@RyanMakiola & @rlt : I this good to go now?

rlt:
@wkande Yes this is good to go now 👍🏻

PR #1127, opened 2022-11-30
Clarify Docker Desktop local provider URL

dcroote:
Follow-up to #1120 as we've identified that using Docker Desktop on any platform requires `http://host.docker.internal:8545`.


wkande (approved):
Thanks, look good.

PR #1133, opened 2022-12-02
made changes related to ois v1.3

metobom:
`operation` is optional in OIS v1.3.0 See [here](https://github.com/api3dao/ois/releases/tag/v1.3.0).


wkande:
@metobom So we have a little issue here. OIS `v1.3` is not the latest release of OIS since `v0.10` of Airnode has not been released. In fact OIS `v1.3` cannot be published until then. Not your fault as no one told you about these things.

I will fix all this for you after I merge this PR today or tomorrow. In the future (because no one told you yet), when a new version of OIS or Airnode is needed please let me know and I will set it up for you. After it gets set up, then you can update it. Sorry no one informed you of this process.

wkande (approved):
Will merge and follow-up with changes to make this version inactive until it is deployed with Airnode `v0.10`.

docs/ois/v1.3/ois.md
```diff
@@ -0,0 +1,611 @@
...
+### 5.1. `name`
+
+(Required) The name of the Airnode endpoint, must be unique in OIS.
+
+<!--OAS equivalent: `paths.{path}.{method}.operationId` of the corresponding
+operation.-->
+
+### 5.2. `operation`
+
+(Optional) An object that refers to an API operation defined in
```
Siegrift, line 392: It would be nice to do this in two commits. First copy the v1.2 docs and in second apply the changes... This way it's hard to review what has changed (you need to compare the 1.2 and 1.3 files in the editor).

Siegrift, line 392: I also wonder if this should be marked as `Optional` as it's not entirely true. It can be optional under special circumstances, but otherwise it's still required. I'd maybe prefer an info or warning section for this. WDYT @wkande @metobom ?

wkande: I believe we have other parameters like this.

See here: [here](https://docs.api3.org/ois/v1.2/ois.html#_4-3-3-in)

wkande: Yes: in the future a new version of OIS should be establish as a base copy of the previous release. Then it should be edited. Normally I do those when asked. I think I need to communicate that to the team again as we have new members so that is on me.

metobom: Sorry about that. Will do it next time.

PR #1137, opened 2022-12-06
Fix byog QRNG

bbenligiray:


docs/qrng/reference/providers.md
```diff
@@ -71,10 +78,16 @@ Australia. Visit their website at
...
 
-## BYOG Random Numbers
+## byog Random Numbers
+
+[byog](https://byog.io/) is an independent group within the API3 ecosystem that
+builds high-impact oracle services.
+
+::: warning Supported chains
+
+byog Random Numbers emulates the QRNG service on [testnets](./chains.md) using [pseudorandom number generation](../README.md).
```
bbenligiray, line 88: Otherwise it might look deceptive

docs/qrng/reference/providers.md
```diff
@@ -39,6 +39,13 @@ Australia. Visit their website at
 
 <!-- Need css for mobile -->
 
+::: warning Supported chains
+
+ANU Quantum Random Numbers is only available on [mainnets](./chains.md).
```
bbenligiray, line 44: Without this you're asking for the user to try to use it on testnets

wkande: So this is a change in that ANU cannot be used on a testnet any longer? So if a user wants to try QRNG they must use a byyog contract address only on testnet.

Can byog use main net?

bbenligiray: Previously
ANU: all mainnets and goerli
Now
ANU: all mainnets (and no testnets)
byog: all testnets (and no mainnets)

metobom (approved):

wkande:
In addition to these changes I will follow-up with a very brief note (of the same) on the `chains.md` page.

PR #1138, opened 2022-12-07
Added small explanation for operation in OIS v1.3

metobom:
`operation` is not required if the API call will be skipped. Added an explanation about it.


wkande (approved):
Merging this change. I think we still need an short explanation (maybe a brief use case) as to why one would want to skip the API call.

PR #1139, opened 2022-12-08
fix: remove second underscore from _times reserved parameter

dcroote:


wkande (approved):
Funny how those type of things lurk about. Good catch.

PR #1148, opened 2022-12-12
Add --cache property to prettier

Siegrift:
Relates to https://github.com/api3dao/airnode/issues/1595

This greatly reduced the prettier execution time. See the screen shot.

image


wkande (approved):
I ran this against the newer VitePress docs and for the command `yarn format` and `yarn format:check` there is a major improvement. Did not seems to matter with the vite build though. Not sure if is using prettier at this point.

Siegrift:
> Did not seems to matter with the vite build though. Not sure if is using prettier at this point.

I don't think so.

> I ran this against the newer VitePress docs and for the command yarn format and yarn format:check there is a major improvement.

Feel free to add this to the new docs repo as well. I didn't want to do it for all our repos.

PR #1152, opened 2022-12-16
fix(ci): update a few links and exclude pokt provider url

dcroote:


wkande (approved):
These little things seemingly live forever don't they? Thanks

PR #1161, opened 2023-01-10
Add airnode-deployer `fetch-files` command

### LINKED ISSUE #1159: Allow users to fetch deployment files from the cloud provider
Documentation for https://github.com/api3dao/airnode/issues/1405

amarthadan:
Close #1159

### DIFF
--- a/docs/airnode/v0.10/grp-providers/docker/deployer-image.md
+++ b/docs/airnode/v0.10/grp-providers/docker/deployer-image.md
@@ -216,7 +216,7 @@ from the [list](./deployer-image.md#list) command above to request information
 about a specific deployment. The retrieved information include deployment's
 Airnode address, stage, Airnode version and the update history. Files for cloud
 provider authentication are needed for the command to run correctly: `aws.env`
-(for AWS) and/or `gcp.json` (for GCP).
+(for AWS) or `gcp.json` (for GCP).
 
 :::: tabs
 
@@ -257,8 +257,7 @@ specify a version of the deployment for which the files should be downloaded. By
 default, the archive with the files is stored in the `config` directory. You can
 change the output directory by providing an `--output-dir` option specifying a
 different directory instead. Files for cloud provider authentication are needed
-for the command to run correctly: `aws.env` (for AWS) and/or `gcp.json` (for
-GCP).
+for the command to run correctly: `aws.env` (for AWS) or `gcp.json` (for GCP).
 
 :::: tabs
 
@@ -293,7 +292,7 @@ Airnode, use the deployment ID from the [list](./deployer-image.md#list) command
 above. Airnode's update history, that can be seen by the
 [info](./deployer-image.md#info) command, will be removed as well. Files for
 cloud provider authentication are needed for the command to run correctly:
-`aws.env` (for AWS) and/or `gcp.json` (for GCP). This is the recommended way to
+`aws.env` (for AWS) or `gcp.json` (for GCP). This is the recommended way to
 remove a deployment, but there are alternatives as described below.
 
 :::: tabs
@@ -327,8 +326,8 @@ When an Airnode was deployed using the `deploy` command, a `receipt.json` file
 was created. This file is used to remove the Airnode. The
 [remove-with-receipt](../../reference/packages/deployer.md#remove-with-receipt)
 command is identical for AWS and GCP. Files for cloud provider authentication
-are needed for the command to run correctly: `aws.env` (for AWS) and/or
-`gcp.json` (for GCP).
+are needed for the command to run correctly: `aws.env` (for AWS) or `gcp.json`
+(for GCP).
 
 :::: tabs
 
--- a/docs/airnode/v0.10/grp-providers/docker/deployer-image.md
+++ b/docs/airnode/v0.10/grp-providers/docker/deployer-image.md
@@ -254,10 +254,13 @@ which should be the files retrieved. By default, the files from the latest
 version of this deployment are fetched. Additionally, you can provide a
 deployment version ID from the [info](./deployer-image.md#info) command above to
 specify a version of the deployment for which the files should be downloaded. By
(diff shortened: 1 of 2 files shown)


dcroote:
Thanks - will test & review tonight

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
```diff
@@ -242,6 +243,48 @@ docker run -it --rm ^
...
 ::::
 
+### `fetch-files`
+
+During the Airnode deployment, your `config.json` and `secrets.env` are uploaded
+to the cloud provider of your choosing. You can use the
+[fetch-files](../../reference/packages/deployer.md#fetch-files) command to
+retrieve them. You need to provide the deployment ID from the
+[list](./deployer-image.md#list) command above to specify the deployment for
+which should be the files retrieved. By default, the files from the latest
```
dcroote: Typo: "for which should be the files retrieved"

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
```diff
@@ -242,6 +243,48 @@ docker run -it --rm ^
...
+version of this deployment are fetched. Additionally, you can provide a
```
dcroote: Perhaps "Alternatively" rather than "Additionally" as providing a deployment version ID is an alternative to the default.

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
```diff
@@ -242,6 +243,48 @@ docker run -it --rm ^
...
+deployment version ID from the [info](./deployer-image.md#info) command above to
+specify a version of the deployment for which the files should be downloaded. By
+default, the archive with the files is stored in the `config` directory. You can
+change the output directory by providing an `--output-dir` option specifying a
+different directory instead. Files for cloud provider authentication are needed
+for the command to run correctly: `aws.env` (for AWS) and/or `gcp.json` (for
```
dcroote: I think just "or" rather than "and/or"

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
```diff
@@ -242,6 +243,48 @@ docker run -it --rm ^
...
+During the Airnode deployment, your `config.json` and `secrets.env` are uploaded
+to the cloud provider of your choosing. You can use the
+[fetch-files](../../reference/packages/deployer.md#fetch-files) command to
+retrieve them. You need to provide the deployment ID from the
+[list](./deployer-image.md#list) command above to specify the deployment for
+which should be the files retrieved. By default, the files from the latest
+version of this deployment are fetched. Additionally, you can provide a
+deployment version ID from the [info](./deployer-image.md#info) command above to
+specify a version of the deployment for which the files should be downloaded. By
+default, the archive with the files is stored in the `config` directory. You can
```
dcroote: I tested the `fetch-files` command and the zip archive got saved to the directory from which I ran the command (`$(pwd)`). Therefore, I would revise this sentence since it pertains to the `config` directory **within the container** and not where the user will actually find the archive.

dcroote (changes requested):
Looks good overall, I suggest a couple minor language fixes and one slightly more substantive change regarding the output location of the command.

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
amarthadan: I used additionally because you're adding one more argument.
`airnode-deployer fetch-files aws2c6ef2b3 3580a278`
instead of
`airnode-deployer fetch-files aws2c6ef2b3`.
Or maybe it would read better as
`Alternatively, you can additionally provide a...`
Or is that too much? :smile:

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
amarthadan: Fair point. Especially since this is the page about the deployer Docker image. Will change.

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
amarthadan: I've copied it from the commands above. But it actually makes sense only for the `list` command as that allows you to list deployment from both cloud providers at the same time (hence you can provide authentication for both). I'll fix it for the rest of the commands.

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
amarthadan: I'm probably just blind but what is the typo? :sweat_smile:

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
dcroote: ~~Ah for both at the same time- ok nice, never mind then.~~ Yes, right, fetching would be for a single provider at a time, I saw your change 👍

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
dcroote: I think that could work, but it's a very small nit so will leave up to you

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
dcroote: Reads well with your update 👍

docs/airnode/v0.10/grp-providers/docker/deployer-image.md
dcroote: Reworded, I think it should be "for which the files should be retrieved" but a simplified version might just be adjusting earlier in the sentence "...command above to specify the desired deployment."

dcroote (approved):
Changes look good; the other very minor discussion points are up to you

wkande:
Deployed

PR #1165, opened 2023-01-17
Replace operations S3 chains source with operations api

### LINKED ISSUE #1166: Replace operations S3 chains source with operations API

vponline:
@wkande This will replace the old (operations) source for chains with the new operations API endpoint. 

There are a lot more chains here and some of them don't have all of the deployment data so perhaps we should wait on merging this until everything is filled (or alternatively filter them out for now). I'll get back on this point.

Closes #1166

### DIFF
--- a/archived/components/beacons/browser2/BeaconList2.vue
+++ b/archived/components/beacons/browser2/BeaconList2.vue
@@ -95,7 +95,7 @@ export default {
           'https://operations-development.s3.amazonaws.com/latest/apis.json'
         );
         const responseChains = await axios.get(
-          'https://operations-development.s3.amazonaws.com/latest/chains.json '
+          'https://db-api-prod.api3.org/api/docs-chains-reference'
         );
         const providers = response.data;
         const chains = responseChains.data;
--- a/docs/.vuepress/components/dapis/browsers/DapiList.vue
+++ b/docs/.vuepress/components/dapis/browsers/DapiList.vue
@@ -97,7 +97,7 @@ export default {
   methods: {
     async init() {
       const responseChains = await axios.get(
-        'https://operations-development.s3.amazonaws.com/latest/chains.json '
+        'https://db-api-prod.api3.org/api/docs-chains-reference'
       );
       this.chains = responseChains.data;
       this.chains = this.sortChainsByName(this.chains);
--- a/docs/.vuepress/components/dapis/chains/ChainsList.vue
+++ b/docs/.vuepress/components/dapis/chains/ChainsList.vue
@@ -46,7 +46,7 @@ export default {
     async loadChainsFromRepo() {
       try {
         const response = await axios.get(
-          'https://operations-development.s3.amazonaws.com/latest/chains.json '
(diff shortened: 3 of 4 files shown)


wkande:
@vponline When you are ready let's plan on going over this one in detail. I need to understand because so I can update the newer docs. This script puzzles me:
```
"chains:build": "curl https://db-api-prod.api3.org/api/docs-chains-reference -o ./docs/.vuepress/chains.json",
```

I use `/docs/.vuepress/chains.json` as a lookup value in many places. So as long as the JSON structure does not change much. Otherwise I just need to understand so I can correct any references I make to the file.

vponline:
> @vponline When you are ready let's plan on going over this one in detail. I need to understand because so I can update the newer docs. This script puzzles me:
> 
> ```
> "chains:build": "curl https://db-api-prod.api3.org/api/docs-chains-reference -o ./docs/.vuepress/chains.json",
> ```
> 
> I use `/docs/.vuepress/chains.json` as a lookup value in many places. So as long as the JSON structure does not change much. Otherwise I just need to understand so I can correct any references I make to the file.

Yes absolutely. Just to clarify, this URL change should not affect you in any way. The content and structure should be the same (assuming I got it right), with the exception of there being more chains available since new deployments have been made. 

Previously, the (now outdated) operations repository generated a JSON file which you were fetching from the AWS S3 bucket with the old URL. This new URL is providing that same data from the (production) operations database that we have been working on for the past few months.

wkande:
@vponline - Yes now I see what you have done to the operations. Two things:

1. The new data structure breaks the Vue component  that is displays **[here](https://api3-docs--pr1165-chains-source-url-szixhiax.web.app/dapis/reference/chains.html)**. I will fix this after merging this PR.

2. Still I do not understand the script you added to `packages.json`? What is the purpose? This is a reference list. Are you trying to keep it current? Will it contain the same formatting and the same fields?

vponline:
> @vponline - Yes now I see what you have done to the operations. Two things:
> 1. The new data structure breaks the Vue component  that is displays **[here](https://api3-docs--pr1165-chains-source-url-szixhiax.web.app/dapis/reference/chains.html)**. I will fix this after merging this PR.

We could also change the data that the endpoint sends you if you don't need all of it. This endpoint was created only for the docs here, so we can make it anything you need.

> 2. Still I do not understand the script you added to `packages.json`? 

Hmm the script was already there, I only changed the URL. If you run it, it will replace get the data from the endpoint and save it to the `chains.json` file (replacing the old data).

> What is the purpose? This is a reference list. Are you trying to keep it current? Will it contain the same formatting and the same fields?

The idea was to keep the docs `chains.json` up to date with supported chains/deployments without requiring you to manually enter chainIds and names.

Currently, the content of the new endpoint is the same as the old, but I can see that `chains.json` has only the `id` and `name`. Are these the only things needed? Again, we can adjust the data that the endpoint returns as needed so you wouldn't need to filter in code.

wkande:
@vponline - Thanks for the clarity. I will work on this this morning and see if the data structure of your operation works. I think it will.

PR #1168, opened 2023-01-18
Describe requester passing reserved parameter value

### LINKED ISSUE #1147: Describe requester passing reserved parameter value
It is currently not described that if neither `default` nor `fixed` are specified for a reserved parameter, the user (requester) is expected to pass the value as a parameter. See the relevant code block in the api3dao/ois repo:

https://github.com/api3dao/ois/blob/20192c541d152e9a90373e6af509656f72280ca4/src/ois.ts#L72-L79

dcroote:
Closes #1147

### DIFF
--- a/docs/ois/v1.4/ois.md
+++ b/docs/ois/v1.4/ois.md
@@ -459,6 +459,12 @@ Each object has the following elements:
 - `fixed`
 - `default`
 
+Note that if a reserved parameter object does not include a `fixed` value, a
+requester can, or may be expected to, supply this value as a parameter. A value
+supplied by a requester overrides a `default`. See the
+[\_gasPrice reserved parameter description](./reserved-parameters.md#gasprice)
+for an example.
+
 #### 5.4.1. `name`
 
 (Required) The name of the reserved parameter. Always starts with `_`.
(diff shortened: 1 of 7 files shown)


docs/ois/v1.4/ois.md
```diff
@@ -473,8 +479,8 @@ If `fixed` is used, `default` has no effect and should not be used.
 #### 5.4.3. `default`
 
 (Optional) The default value for the reserved parameter. Used when no value is
-provided. If `default` is used, `fixed` must not be used because it would
-override the default value.
+provided by a requester. If `default` is used, `fixed` must not be used because
```
Siegrift: fyi: validator validates this, so I'd rephrase it to: `Only one of the default or fixed can be used.`

Siegrift: or maybe: `Cannot be used together with "fixed"` and similar message for the `fixed`.

Siegrift (approved):
👍 LGTM

docs/ois/v1.4/ois.md
dcroote: Good call 👍  I updated the language for previous versions as well since this validation was present when OIS became a standalone package (v1.0.0)

PR #1171, opened 2023-01-26
20230122 fix dapis chain dsource

wkande:
@vponline created a new endpoint for the dAPI chain list. Going forward changes to this list can be made at the backend. As such a lot of code was removed from the Vue components that calls the endpoint and displays the chain list.


wkande:
I did slide in some other commits related to simple cleanup in Airnode `v0.10`.

Issue #1130, opened 2022-12-01
Add a new reserved parameter to allow minConfirmations to be specified


amarthadan:
Documentation for https://github.com/api3dao/airnode/issues/1393
One of the TODOs from the Migration guide should be part of this issue (https://github.com/api3dao/api3-docs/pull/1174#issuecomment-1406386840)

wkande:
I updated the migration guide TODO for **_Skip the API call_**  in PR#1118

Issue #1118, opened 2022-11-21
Allow API call to be skipped in endpoints


amarthadan:
Documentation for https://github.com/api3dao/airnode/issues/1527
One of the TODOs from the Migration guide should be part of this issue (https://github.com/api3dao/api3-docs/pull/1174#issuecomment-1406386840)

wkande:
I think we still need a short explanation (maybe a brief use case) as to why one would want to skip the API call.

wkande:
@aTeoke , @metobom - I am working on this one today. Adding myself as an assignee.

wkande:
@metobom and @aTeoke 
Please check this review at these preview links or in the code.

- [5.2. operation](https://api3-docs--pr1179-20230131-1118-ois-sk-a5t5tufj.web.app/ois/v1.4/ois.html#_5-2-operation)
- [Skip the API provider call](https://api3-docs--pr1179-20230131-1118-ois-sk-a5t5tufj.web.app/ois/v1.4/processing.html#skip-the-api-provider-call)

metobom:
> @metobom and @aTeoke Please check this review at these preview links or in the code.
> 
> * [5.2. operation](https://api3-docs--pr1179-20230131-1118-ois-sk-a5t5tufj.web.app/ois/v1.4/ois.html#_5-2-operation)
> * [Skip the API provider call](https://api3-docs--pr1179-20230131-1118-ois-sk-a5t5tufj.web.app/ois/v1.4/processing.html#skip-the-api-provider-call)

* Is `API provider call` the right term? Should it be just `API call`? 
* I did not think of it but, in the random number generation example output will always be zero since `Math.random()` returns between 0 and 1 and `_type` in `reservedParameters` is `uint256`. If that's a problem, it can be changed to `Math.floor(Math.random() * 100)` to make it return between 0 and 100.

aTeoke:
Let's change the example please.

wkande:
> Is API provider call the right term? Should it be just API call?

I am fine either way:

- (Required: when calling an API provider operation)      **> becomes >**     (Required: when calling an API)
- Instead of calling an API provider operation     **> becomes >**    Instead of calling an API

wkande:
@amarthadan - I updated the Airnode `v0.10` migrations guide with regards to skipping an API call. It will show up in the next push to this PR.

wkande:
[...]

PR #1174, opened 2023-01-27
fix: minor updates to v0.10 migration guide

dcroote:
One addition, and otherwise minor language / link updates

### DIFF
--- a/docs/airnode/v0.10/reference/migration.md
+++ b/docs/airnode/v0.10/reference/migration.md
@@ -41,8 +41,6 @@ such as airnode-deployer, airnode-admin, etc., and new features.
 6. Command `airnode-deployer remove-with-deployment-details` no longer exists.
    It was replaced by a command `airnode-deployer remove <deployment ID>`.
 
-7. Integrations within `airnode-examples` default to `DEBUG` log level.
-
 ### Details
 
 1. `ois[n].oisFormat`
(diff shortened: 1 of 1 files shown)


docs/airnode/v0.10/reference/migration.md
```diff
@@ -41,6 +41,8 @@ Airnode, such as airnode-deployer, airnode-admin, etc., and new features.
 6. Command `airnode-deployer remove-with-deployment-details` no longer exists.
    It was replaced by a command `airnode-deployer remove <deployment ID>`.
 
+7. Integrations within `airnode-examples` default to `DEBUG` log level.
```
amarthadan: I knew about this just haven't considered it worth mentioning :smile: 
But if you want to mention it, I would move it to the new feature section. The migration guide should contain only the things necessary for you to change when you're migrating from the 0.9.x config to 0.10.x config. And this change was internal and doesn't require any user action.

amarthadan (approved):
Just a comment regarding the new migration point. Other than that LGTM. Thanks for the fixes :pray:

docs/airnode/v0.10/reference/migration.md
dcroote: Agreed it's very minor and good point about location. Since there is no harm in keeping it and already typed up, I moved it.

wkande:
Will there be another PR then for the TODOs? Or should I wait for changes to this PR?

---
**New features**

Skipping API calls to just run pre/post processing (**TODO** https://github.com/api3dao/api3-docs/issues/1118 still open)
There are two new reserved parameters available, [_gasPrice](https://github.com/api3dao/api3-docs/blob/8d493faf7175e904c17c28a76ef9e9beb38f0087/docs/ois/v1.4/reserved-parameters.html#gasprice) and _minConfirmations. You can read more about how and what are they used for in the [documentation](https://github.com/api3dao/api3-docs/blob/8d493faf7175e904c17c28a76ef9e9beb38f0087/docs/ois/v1.4/reserved-parameters.html). (**TODO** link documentation for _minConfirmations once it's done https://github.com/api3dao/api3-docs/issues/1130)

amarthadan:
Up to you and @dcroote but my intention was to address those in separate PRs.

dcroote:
> Up to you and @dcroote but my intention was to address those in separate PRs.

Agreed

wkande:
Thanks, I will merge this later today with my other PRs.

PR #1177, opened 2023-01-30
feat: describe new _minConfirmations reserved parameter

### LINKED ISSUE #1130: Add a new reserved parameter to allow minConfirmations to be specified

dcroote:
Closes #1130

### DIFF
--- a/docs/airnode/v0.10/grp-providers/guides/build-an-airnode/api-integration.md
+++ b/docs/airnode/v0.10/grp-providers/guides/build-an-airnode/api-integration.md
@@ -395,9 +395,10 @@ start with an underscore. See the
 [related OIS docs](/ois/v1.4/ois.md#_5-4-reservedparameters) for more
 information.
 
-The current list of reserved parameters are `_type`, `_path`, `_times`, and
-`_gasPrice`. See [Reserved Parameters](/ois/v1.4/reserved-parameters.md) in the
-OIS document set to understand what each of these parameters are for.
+The current list of reserved parameters are `_type`, `_path`, `_times`,
+`_gasPrice`, and `_minConfirmations`. See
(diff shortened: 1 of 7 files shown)


wkande:
Somewhere between `v0.9` and `v0.10` the sentence below (strikethrough) was removed for. I see it is also gone from this PR. Is this intended? The content below is from `v0.9`, of course there are now 4 reserved parameters available for `v0.10`.

> The current list of reserved parameters are `_type`, `_path` and `_times`. See [Reserved Parameters](/reference/ois/latest/reserved-parameters.md) in the OIS document set to understand what each of these parameters are for. In most cases, all three should be defined as reserved parameters with no fixed/default values, as doing so provides the requester with the most flexibility. ~~In most cases, all three should be defined as reserved parameters with no fixed/default values, as doing so provides the requester with the most flexibility.~~

dcroote:
I removed that sentence because otherwise users would be effectively allowed to call any endpoint (through an arbitrary `_path`) and need to know the what type of response it returns and if numerical, what to multiply it by. We don't have those 3 reserved parameters empty in an tutorial, example, etc.

This PR will make 5 reserved parameters. 

Converted to draft because the request ordering logic needs to change.

wkande:
I have migrated this PR to the point that Derek will make new changes mentioned above.

wkande:
@dcroote I see this was changed back to Open from Draft. Can I merge it?

dcroote:
Yes and same with the corresponding Vitepress PR

PR #1193, opened 2023-02-08
Add Airnode v0.11 migration guide

dcroote:
This is a draft / WIP branch that should be added to during Airnode v0.11 development.


wkande:
If possible I would like to discuss the formatting of this document. The content is good but the heading2 and heading3  labels do not flow well when the TOC is added back into this document. 

<h1>Migration Guide</h1>
... content
<h2>Summary</h2>
... content
<h2>Details</h2>
... content
<h2>New features</h2>
... content

---
Basically this document is short and does not need nested heading 3 elements. Also the existing "Migration" heading2 element does not have content and is simply a repeat of heading1 in a shortened form. As I look at this PR is appears that @dcroote has already implemented this? If so then we are on the same page.

dcroote:
@wkande - I think the heading 3 nesting is useful to distinguish which sections correspond to `config.json` changes (summary and details) and which don't (new features), but I agree it's currently not clear e.g. in the [v0.10 migration guide](https://docs.api3.org/airnode/v0.10/reference/migration.html). I propose changing to:

## Migrating `config.json`
### Summary
### Details
## New v0.X features

amarthadan:
Is there a reason why is this a draft PR instead of having it already in the 0.11 directory like the rest of the docs?

dcroote:
> Is there a reason why is this a draft PR instead of having it already in the 0.11 directory like the rest of the docs?

My intention with draft status was so that folks could contribute and it wouldn't be merged prematurely, but so long as we know not to merge until everything is complete then it's fine as a regular PR.

amarthadan:
> My intention with draft status was so that folks could contribute and it wouldn't be merged prematurely, but so long as we know not to merge until everything is complete then it's fine as a regular PR.

No, I meant why not merge it? It would be easier to make changes there when we're adding stuff to the docs. This way we first update the docs with one PR and then need to add a commit to this one. Even though it's both in the `0.11` directory that is essentially hidden.

dcroote:
Ah ok, yes sure will merge

wkande:
@dcroote - 
> The Node.js version of Airnode was upgraded from 14 to 18.

The example code below this statement just uses `"environment": "Node",` for pre/post processing. Does this mean the Node version is picked up from the config file or somewhere else?

dcroote:
The Node version of Airnode is set in `package.json` (i.e. when it's built) here: https://github.com/api3dao/airnode/blob/acaea82a60370f7df5b8ab3e0b6adbffae4ed114/package.json#L7. Do you think there is a better phrasing then to avoid confusion?

wkande:
Then I think it is OK. So now "environment": "Node", just declares the language used.

PR #1220, opened 2023-03-14
OEV gateway

### LINKED ISSUE #1215: Implement the HTTP gateway to facilitate OEV auctions
Documentation for https://github.com/api3dao/airnode/issues/1579

amarthadan:
Close #1215

I've added a separate page for the OEV gateway as the purpose of this gateway is quite different than the rest of the gateways. But feel free to rearrange it as you see fit if you feel differently.

Jacob (can't tag you for some reason, will DM you) feel free to add more details but it feels like it's not really necessary on the Airnode side and anything else important will be part of the OEV relay docs.

### DIFF
--- a/docs/airnode/v0.11/grp-providers/guides/build-an-airnode/oev-gateway.md
+++ b/docs/airnode/v0.11/grp-providers/guides/build-an-airnode/oev-gateway.md
@@ -1,0 +1,172 @@
+---
+title: OEV Gateway (optional)
+docSetName: Airnode v0.11
+folder: API Providers > Build an Airnode
+basePath: /airnode/v0.11
+tags:
+---
+
+<TitleSpan>{{$frontmatter.folder}}</TitleSpan>
+
+# {{$frontmatter.title}}
+
+<VersionWarning/>
+
+<TocHeader />
+<TOC class="table-of-contents" :include-level="[2,3]" />
+
+OEV gateway is used in the OEV flow to sign the data won in the auction. The
+data is signed by the Airnode so that only the searcher who won the auction can
+use it to update the data feed.
+
+## Setup
+
+Enable the gateway in the `config.json` file via `nodeSettings.oevGateway`
+section.
+
+- **enabled**: A boolean to enable/disable for the gateway.
+- **maxConcurrency**: (optional) A number higher than zero that represents the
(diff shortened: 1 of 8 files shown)


jgcogsystematictrading:
It looks fine, my only thought was maybe you could expand on what is included in the returned signature from the gateway but probably not be too important.

docs/airnode/v0.11/grp-providers/guides/build-an-airnode/oev-gateway.md
```diff
@@ -0,0 +1,170 @@
...
+- `bidderAddress` - Blockchain address of the winning searcher.
+- `bidAmount` - Bid amount that won the auction.
+- `signedData` - A list of beacon data to be signed. It can contain two types:
+  full beacon data, only beacon metadata
+- `airnodeAddress` - Airnode address identifying a beacon
+- `endpointId` - Endpoint ID identifying a beacon
+- `encodedParameters` - Parameters in their encoded form identifying a beacon
+- `timestamp` - UNIX timestamp of the beacon data
+- `encodedValue` - Beacon value in its encoded form
+- `signature` - Signature of the beacon data
```
dcroote: I would indent these bullets given they belong within `signedData`.
```suggestion
  - `airnodeAddress` - Airnode address identifying a beacon
  - `endpointId` - Endpoint ID identifying a beacon
  - `encodedParameters` - Parameters in their encoded form identifying a beacon
  - `timestamp` - UNIX timestamp of the beacon data
  - `encodedValue` - Beacon value in its encoded form
  - `signature` - Signature of the beacon data
```

dcroote (approved):
LGTM, one minor suggestion

wkande:
@dcroote @amarthadan Should this be hidden until OEV is actually functional, or maybe it is. It is a feature that by having presence in the docs, readers might believe is something that can implement. There are no OEV auctions right now if I am not mistaken. Do I not understand?

amarthadan:
You are correct, there are no auctions at the moment. But this is not really a user-facing piece anyway. I think it's ok to include it, it's an Airnode feature. But nobody will be actually able to use it really :man_shrugging:. @jgcogsystematictrading WDYT?

PR #1243, opened 2023-12-27
fix: remove use of references.json chainNames field

### LINKED ISSUE #1242: Remove use of chainNames and networks from references.json
https://github.com/api3dao/airnode/issues/1926#issue-2056537249

dcroote:
Closes #1242. Since these old docs are deprecated, I've opted for skipping the display of chains that aren't present in [chains.json](https://github.com/api3dao/api3-docs/blob/main/docs/.vuepress/chains.json) rather than invest in upgrading these docs to use @api3/chains.

### DIFF
--- a/docs/dapis/reference/chains.md
+++ b/docs/dapis/reference/chains.md
@@ -7,9 +7,6 @@ folder: Reference
 
 # {{$frontmatter.title}}
 
-dAPIs can be read on the following chains using the
-[DapiServer](../#dapiserver-sol) contract addresses listed below. Use the
-[API3 Market](https://market.api3.org/) website to find the desired dAPI names
-and the networks they are available on.
-
-<dapis-chains-ChainsList/>
+Please refer to the new API3 documentation's
+[dAPIs Reference](https://docs.api3.org/reference/dapis/chains/) for the latest
+technical details on accessing dAPIs.
--- a/docs/dapis/reference/chains.md
+++ b/docs/dapis/reference/chains.md
@@ -7,6 +7,6 @@ folder: Reference
 
 # {{$frontmatter.title}}
 
-Please refer to the new API3 documentation's
-[dAPIs Reference](https://docs.api3.org/reference/dapis/chains/) for the latest
-technical details on accessing dAPIs.
+Please refer to the new API3 documentation for the latest
+[Chains and Contracts](https://docs.api3.org/reference/dapis/chains/) page.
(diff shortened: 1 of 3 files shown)


wkande:
Your logic makes sense to me for the Airnode contract addresses.

There may be another issue here with the dAPI **Chain and Contracts** page: https://old-docs.api3.org/dapis/reference/chains.html

1. There are chains without contract addresses, meaning they may not be relevant chains. This may be due to the endpoint returning more data as time has progressed.
2. It is using an endpoint that may go away
3. It uses `DapiServer` and not `Api3ServerV1` and `ProxyFactory`

I guess the data is part of what dAPIs were back then but not now. But that endpoint probably needs to go. Let me propose this idea which is somewhat laziness. Since reading dAPIs is not possible the way the old docs describe we provide a  link back to the new docs **Chains and Contracts** page? 

Having said that a lot of what is in the old docs for dAPIs is legacy in that the user really needs to move on to the new docs to use dAPIs. But at the very least the **Chain and Contracts** page is very problematic I think. 

Another idea is to stop using the endpoint and create a markdown table of what was the data at the time. Probably the better solution

dcroote:
> But that endpoint probably needs to go. Let me propose this idea which is somewhat laziness. Since reading dAPIs is not possible the way the old docs describe we provide a link back to the new docs Chains and Contracts page?

Good thought- I agree it's best to refer to the new docs and I've done so in the latest commit- take a look and let me know what you think.

docs/dapis/reference/chains.md
```diff
@@ -7,9 +7,6 @@ folder: Reference
...
 # {{$frontmatter.title}}
 
-dAPIs can be read on the following chains using the
-[DapiServer](../#dapiserver-sol) contract addresses listed below. Use the
-[API3 Market](https://market.api3.org/) website to find the desired dAPI names
-and the networks they are available on.
-
-<dapis-chains-ChainsList/>
+Please refer to the new API3 documentation's
+[dAPIs Reference](https://docs.api3.org/reference/dapis/chains/) for the latest
```
wkande: This idea works best then. But are we not primarily sending them to the latest list of the Chains and Contracts page?

> Please refer to the new API3 documentation for the latest [Chains and Contracts](https://docs.api3.org/reference/dapis/chains/) page. There you will also find the current   technical information on accessing dAPIs.

Something like that.

wkande (approved):
LGTM

